PR #7, opened 2016-12-29
Add a load balancer matching across the local host, then round-robin

ddobrin:
Let's discuss it first. Code is there and can be merged afterwards if it matches the need


stevehu:
Thanks Dan.

PR #56, opened 2017-05-02
test case need to be fix

chenyan71:


stevehu:
As we have discussed, this is due to the format of service.yml. No need to merge.

PR #92, opened 2017-08-23
Update README.md

joaozitopolo:
There are a typo at word "Angular"


stevehu:
@joaozitopolo Nice catch! I should have focused more on documentation but time is very limited. Thanks a lot for your help.

PR #152, opened 2018-01-09
Fix HTTP verbs

morganseznec:
Consul 1.0.0 started enforcing HTTP verbs in many endpoints. GET methods must be changed to PUT.
See: https://github.com/hashicorp/consul/blob/v1.0.0/CHANGELOG.md


stevehu:
@morganseznec Thanks a lot for the fix. I was shocked when I see this as we have customers using it on production with Consul 0.8.x and 0.9.x and I need to do a lot of regression test to ensure that PUT works with Consul server below 1.0.0 version. I will take this chance to revisit [service registry and discovery toturial](https://doc.networknt.com/tutorial/common/discovery/) and test it with both versions of Consul.

PR #174, opened 2018-03-18
feat(util): Add toByteBuffer overload util method to convert files.

NicholasAzar:
Hey, I think I'm going to need this util function to convert the zip file into a bytebuffer for returning it to the client.

In regards to: https://github.com/networknt/light-codegen/issues/112

### DIFF
--- a/utility/pom.xml
+++ b/utility/pom.xml
@@ -54,6 +54,10 @@
             <groupId>org.owasp.encoder</groupId>
             <artifactId>encoder</artifactId>
         </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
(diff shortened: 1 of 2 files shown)


stevehu:
@NicholasAzar I have switched to a method inside the same class and tried to remove some other dependencies. I goal is the remove all the apache libraries in the future so that our delivery package becomes smaller. Thanks.

NicholasAzar:
@stevehu Oops! didn't see that there.. Thanks!!

stevehu:
For most of APIs, the size is still around 20MB and I want to reduce it to around 10MB. This will make DevOps pipeline more efficient. It is hard and will take a long time. We will get there eventually.

PR #213, opened 2018-06-01
testing addition of graceful shutdown handler

NicholasAzar:
Hey @stevehu ,

I'm sure you've looked into integrating the graceful shutdown handler into l4j before.. was just wondering why it's been omitted... was it for performance reasons? Or is there another reason?

Thanks


stevehu:
Nicely done. We should put this on the server. It is in my mind all the time but never got it implemented.

NicholasAzar:
We certainly need a performance test comparison before merging though. From just looking at the source it could potentially add a fair bit of overhead

PR #218, opened 2018-06-13
Fix216

rpinaa:
Hi, here it is the content type handler to set a default content-type.


stevehu:
@rpinaa Thanks a lot for your help.

PR #219, opened 2018-06-14
Adding InputStream to String conversion utility

NicholasAzar:
Optimizing InputStream to String conversion per performance test conducted here:
https://stackoverflow.com/a/35446009

Including this change in the BodyHandler for now.


stevehu:
Cool. Thanks.

PR #245, opened 2018-08-03
Safer set exchange status

### LINKED ISSUE #244: catastrophic setExchangeStatus calls without args
If `LightHttpHandler.setExchaggeStatus` is called without enough `args` to populate the description template string for the particular state's description, then this raises an exception which eclipses the original problem. This tends to happen when the user (me) is doing something wrong and hitting unexpected and less tested code paths. The user (me) is also already sufficiently confused without the error handler failing.

One example:
```
java.util.MissingFormatArgumentException: Format specifier '%s'
        at java.util.Formatter.format(Formatter.java:2519)
        at java.util.Formatter.format(Formatter.java:2455)
        at java.lang.String.format(String.java:2940)
        at com.networknt.status.Status.<init>(Status.java:89)
        at com.networknt.handler.LightHttpHandler.setExchangeStatus(LightHttpHandler.java:21)
        at com.networknt.handler.OrchestrationHandler.handleRequest(OrchestrationHandler.java:15)
```

I have a local change to `Status` that I can submit as a PR which logs an informative warning and then sets the intended status but leaves the `%s` (or other format specifiers) in the description.

logi:
Fixes #244

### DIFF
--- a/handler/src/main/java/com/networknt/handler/OrchestrationHandler.java
+++ b/handler/src/main/java/com/networknt/handler/OrchestrationHandler.java
@@ -12,7 +12,8 @@ public class OrchestrationHandler implements LightHttpHandler {
         if (Handler.start(exchange)) {
             Handler.next(exchange);
         } else {
-            setExchangeStatus(exchange, "ERR10048");
+            String methodPath = String.format("%s %s", exchange.getRequestMethod(), exchange.getRequestPath());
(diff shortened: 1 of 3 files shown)


stevehu:
Cool. Thanks @logi Someone mentioned the second issue and we are still trying to figure out how to fix it :)

PR #253, opened 2018-08-10
 Descriptive Exception on unknown chain or handler in handler.yml

logi:
Before it would fail with a NullPointerException or return 200 with empty body.


stevehu:
@logi Thanks for your help.

PR #254, opened 2018-08-10
Don't require ignored config

logi:
It seems someone beat me to this fix, but I took it one small step further to not even look for a `HandlerProvider` when it won't be used.


stevehu:
@logi Thanks a lot.

Issue #258, opened 2018-08-13
Support HandlerProvider definitions in handler.yml configuration definitions


NicholasAzar:
Per suggestion in #249 , providing support for the list of handlers configured in `handler.yml` to allow listing of implementations of `HandlerProvider`.

NicholasAzar:
@ddobrin no, nothing is changing in the `handler.yml` definition. Only that the classes being passed in can now implement `HandlerProvider` instead of just `HttpHandler`.

PR #259, opened 2018-08-13
Adding HandlerProvider support to list of handlers.

NicholasAzar:
#258

### DIFF
--- a/handler/src/main/java/com/networknt/handler/Handler.java
+++ b/handler/src/main/java/com/networknt/handler/Handler.java
@@ -215,6 +215,32 @@ public class Handler {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


NicholasAzar:
@logi fyi

ddobrin:
Hi @NicholasAzar,
Is there an example on how HandlerProviders would be added to the handler.yml?

NicholasAzar:
@ddobrin it would be the same way as HttpHandlers without constructors. Nothing new in the `handler.yml` interface definition as part of this change.

ddobrin:
@NicholasAzar : I agree that we should fail fast here, as we wish to support only certain type of handlers (if need be, it can be enhanced on demand later)

Issue #249, opened 2018-08-09
Feedback request to change interface of handler.yml to support multi-path/multi-method combinations


NicholasAzar:
Current interface is designed such that each path in the path list defines a single target url with a single http method (as follows):
```yaml
- path: /v1/health
  method: get
  exec:
      - unsecured
      - health
```

In order to support more terse configurations, we could support multiple items in the path field (changing name to paths), and multiple items in the method field (methods):

```yaml
- paths: 
      - /v1/health
      - /v2/health
  methods:
      - get
      - post
  exec:
      - unsecured
      - health
```

In which case both paths would support both get and post methods (cartesian product).

Any feedback for this would be greatly appreciated.

From: https://gitter.im/networknt/light-4j?at=5b647baba9426c705d210ebc

NicholasAzar:
Paging @stevehu @ddobrin @miklish

stevehu:
I think this is a good idea. It makes the definition more efficient. Since the API is not stabilized yet, it is the good time to make the change. @ddobrin and @logi what do you think? You are the most impacted. Thanks.

ddobrin:
I thinks this could be useful, maybe combine health and server info for brevity under 1 declaration, for ex. 

Now, there are a number of items to be addressed, as listed under #247

I would like for us to:
- **close #247 first**, as it changes 2 or more classes, as well as the status.yml 
- document the interface for the handler.yml
   - for ex, for legibility, I'd prefer - health: com.networknt.health.HealthGetHandler to the existing
                                                        - com.networknt.health.HealthGetHandler@health
- only then make the change for a simplified set of paths.

How does this sound?

ddobrin:
To add to the concern of impact, all APIs in my current project need a certain refactoring after the next light-4j release, therefore the changes will be made if necessary.

After this effort, the interface to the handler.yml needs to be remain backwards compatible

stevehu:
@ddobrin As you have the most APIs implemented, we really need your input to ensure we take the best approach. It would be very hard to make changes down the road. Thanks.

ddobrin:
Logi,
right now they are mutually exclusive; either the backwards compatible mechanism using service.yml or the newer handler.yml.

In the first, you are not specifying the routing handlers, however in the newer one you do.

This will be rationalized

stevehu:
[...]

PR #282, opened 2018-09-07
Related to #249, adds an EndpointSource interface for injecting path,…

logi:
…method pairs into the list of paths at a particular location depending on the implementing class. Intended for injecting endpoints from OpenApi and Swagger spec files but may be useful for the standard graphql endpoints.

### DIFF
--- a/handler/src/main/java/com/networknt/handler/config/PathChain.java
+++ b/handler/src/main/java/com/networknt/handler/config/PathChain.java
@@ -2,15 +2,22 @@ package com.networknt.handler.config;
 
 import com.networknt.utility.NetUtils;
 
+import java.util.ArrayList;
 import java.util.List;
 
 /**
  * @author Nicholas Azar
  */
 public class PathChain {
+
+    private String source;
     private String path;
     private String method;
     private List<String> exec;
+
+    public String getSource() { return source; }
+
+    public void setSource(String source) { this.source = source; }
 
     public String getPath() {
         return path;
@@ -32,11 +39,37 @@ public class PathChain {
         return method;
     }
 
-    public void setMethod(String method) throws Exception {
-        if (NetUtils.METHODS.contains(method.toUpperCase())) {
-            this.method = method;
+    public void setMethod(String method) { this.method = method; }
+
+    @Override
+    public String toString() {
+        if(path != null) {
+            return path + "@" + method + " → " + exec;
         } else {
-            throw new Exception("Invalid HTTP method provided: " + method);
+            return source + "() → " + exec;
+        }
+    }
+    /**
+     * Validate the settings and raise Exception on error.
+     * The origin is used to help locate problems.
+     */
+    public void validate(String origin) throws Exception {
+        List<String> problems = new ArrayList<>();
+        if(source == null) {
+            if(path == null) {
+                problems.add("You must specify either path or source");
+            } else if(method == null) {
+                problems.add("You must specify method along with path: " + path);
+            }
+        } else {
+            if(path != null) problems.add("Conflicting source: " + source + " and path: " + path);
+            if(path != null) problems.add("Conflicting source: " + source + " and method: " + method);
+        }
+        if(method != null && !NetUtils.METHODS.contains(method.toUpperCase())) {
+            problems.add("Invalid HTTP method: " + method);
+        }
+        if(!problems.isEmpty()) {
+            throw new Exception("Bad paths element in " + origin + " [ " + String.join(" | ", problems) + " ]");
         }
     }
 }
--- a/handler/src/main/java/com/networknt/handler/config/PathChain.java
+++ b/handler/src/main/java/com/networknt/handler/config/PathChain.java
@@ -62,8 +62,12 @@ public class PathChain {
                 problems.add("You must specify method along with path: " + path);
             }
         } else {
-            if(path != null) problems.add("Conflicting source: " + source + " and path: " + path);
-            if(path != null) problems.add("Conflicting source: " + source + " and method: " + method);
+            if(path != null) {
+                problems.add("Conflicting source: " + source + " and path: " + path);
+            }
+            if(method != null) {
+                problems.add("Conflicting source: " + source + " and method: " + method);
+            }
         }
         if(method != null && !NetUtils.METHODS.contains(method.toUpperCase())) {
             problems.add("Invalid HTTP method: " + method);
--- a/handler/src/main/java/com/networknt/handler/config/PathChain.java
+++ b/handler/src/main/java/com/networknt/handler/config/PathChain.java
@@ -53,7 +53,7 @@ public class PathChain {
      * Validate the settings and raise Exception on error.
      * The origin is used to help locate problems.
      */
-    public void validate(String origin) throws Exception {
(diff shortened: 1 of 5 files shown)


logi:
Yeah, that's broken. The strangest thing is that it built and ran locally.

handler/src/main/java/com/networknt/handler/config/PathChain.java
```diff
@@ -32,11 +39,37 @@ public String getMethod() {
...
         } else {
-            throw new Exception("Invalid HTTP method provided: " + method);
+            return source + "() → " + exec;
+        }
+    }
+    /**
+     * Validate the settings and raise Exception on error.
+     * The origin is used to help locate problems.
+     */
+    public void validate(String origin) {
```
NicholasAzar: doesn't seem like there's a check against source and path both being null?

ddobrin:
private static void addSourceChain(PathChain sourceChain)

1. sourceChain can be null in: sourcedPath.validate(sourceChain.getSource());
2. Should we maybe throw a FrameworkException, which is a runtime exception as well, yet allows us to create a Status at the point of throw?

private static void addPathChain(PathChain pathChain)
1. Same comment on checking for null

handler/src/main/java/com/networknt/handler/config/PathChain.java
```diff
@@ -32,11 +39,37 @@ public String getMethod() {
...
+        List<String> problems = new ArrayList<>();
+        if(source == null) {
+            if(path == null) {
+                problems.add("You must specify either path or source");
+            } else if(method == null) {
+                problems.add("You must specify method along with path: " + path);
+            }
+        } else {
+            if(path != null) problems.add("Conflicting source: " + source + " and path: " + path);
+            if(path != null) problems.add("Conflicting source: " + source + " and method: " + method);
```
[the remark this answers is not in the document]
logi: No?

I'm trying to run all the check and report all the problems in a single validation call so the deployer doesn't have to go through a build/deploy/run cycle to catch each problem separately. however, the second `path != null` should be `method != null`. I have written a unit test for that case and fixed it locally. Will push shortly.

handler/src/main/java/com/networknt/handler/config/PathChain.java
logi: That would be addressed here:
```java
        if(source == null) {
            if(path == null) {
                problems.add("You must specify either path or source");
```
I have added a unit test for this as well.

handler/src/main/java/com/networknt/handler/config/PathChain.java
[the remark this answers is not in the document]
NicholasAzar: The second `path != null` to `method != null` is what i meant by "check should be method" 😆 

Thanks!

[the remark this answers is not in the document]
logi: The second check there should be `method != null` so that's a bug. It's fixed locally with unit tests and will be pushed in a bit.

I don't think the code would become more readable by extracting those two lines into s method:
```
addProblemsConflictWithSource(problems, source, path, "path");
addProblemsConflictWithSource(problems, source, method, "method");
```

logi:
@ddobrin 
`private static void addSourceChain(PathChain sourceChain)`

1. The `sourceChain` comes from `config.getPaths()` where it is deserialised from the `paths` list in `handler.yml`. It was used without null checks before this change and I'm having a hard time seeing how a null could appear. Possibly by writing:
    ```
    paths:
    - !!null
    ````
    at which point I think the developer should not be surprised to get an NPE. That NPE would actually happen at `pathChain.validate(configName + " config")`. I started writing a null check there and just felt a bit silly when I had written:
    ```java
    if (pathChain==null) {
        throw new NullPointerException("Null path found in handler.yml");
    }
    ```
    since that's not really much more informative.

1. "The FrameworkException will be handled by ExceptionHandler" according to the docs so it is intended for raising inside handler chains in response to a request coming in. However this code is running in `Server.start()` outside request handling.

    I brought up the idea of a more specific `RuntimeException` to throw during srartup with @stevehu in the gitter channel but he was happy just to use an RE.

`private static void addPathChain(PathChain pathChain)`

1. There are now two cases for where the `pathChain` comes from. Either like in `addSourceChain` which is essentially unaffected by this change or the `pathChain` was instantiated inside `addSourceChain` and can't be null.

logi:
@ddobrin I just tried changing `test/resources/handler.yml` file to contain:
```yaml
paths:
  - path: '/test'
    method: 'get'
    exec:
      - secondBeforeFirst
      - third
  - path: !!null
  - path: '/v2/health'
    method: 'post'
    exec:
      - secondBeforeFirst
      - third
```
and ran the unit tests. Then `HandlerTest.validConfig_init_handlersCreated` fails with:
```
java.lang.RuntimeException: Bad paths element in handler config [ You must specify either path or source ]
```
so Jackson tuns that into an uninitialised object rather than null and the problem is identified by the validation.

handler/src/test/java/com/networknt/handler/PathChainTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Test
+    public void validate_Source() {
+        PathChain chain = new PathChain();
+        chain.setSource("a.source.Class");
+        chain.validate("unit test config");
+    }
+
+    @Test
+    public void validate_NeitherPathNorSource() {
```
logi: @NicholasAzar This tests the case where path==null and source==null

handler/src/main/java/com/networknt/handler/config/PathChain.java
NicholasAzar: Not sure how i missed that before 🙈 

Thanks!

stevehu:
@logi Is it possible to post an example of handler.yml so that I can use it as a template for the light-codegen?

logi:
@stevehu there is a small snippet on https://github.com/networknt/light-rest-4j/pull/55 which had to be a separate PR since this is spread over so many repos. I have a super trivial change for light-graphql-4j as well that just returns the standard `/graphql@foo` and `/subscribe@foo` endpoints but there is no PR yet.

Here is my current `paths` section from `handler.yml` in our test project. I've inlined the chains that were only referenced once but that's a matter of taste.
```yaml
paths:

# Infrastructure End-Points

- path: "/usernotes/v1/health"
  method: GET
  exec:
  - com.networknt.health.HealthGetHandler

- path: "/usernotes/v1/server/info"
  method: GET
  exec:
  - com.networknt.info.ServerInfoGetHandler

# Standard GraphQL End-points

- source: com.networknt.graphql.router.GraphqlEndpointSource
  exec:
  - common-chain
  #- graphql-verify  # TODO#8: Requires security tokens
  - audit
  - graphql-validator
  - graphql-handler

# End-Points From openapi.json

- source: com.networknt.openapi.OpenApiEndpointSource
  exec:
  - common-chain
  - openapi-parser
  #- openapi-verify  # TODO#8: Requires security tokens
  - body
  - audit
  - sanitizer
  - openapi-validator
  - rest-handler
```

stevehu:
@logi Thanks for the example. It gives me a lot of insight into how this works.

PR #292, opened 2018-09-15
Updated oauth helper files to handle SAMLBearer grant type

dguncb:
We have a new requirement to make a SAML2Bearer grant type request to the oauth server. In this design the client will also present a JWT token. The changes here relate to the oauth help files. There will be a new handler in the light router .

### DIFF
--- a/client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
+++ b/client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2016 Network New Technologies Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.networknt.client.oauth;
+
+
+import com.networknt.client.Http2Client;
+import com.networknt.common.DecryptUtil;
+import com.networknt.common.SecretConstants;
+import com.networknt.config.Config;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This holds values used to call the SAML Bearer grant flow from the OAuth Server.
+ * In this version client presents a SAML token and a JWT token (requirment for project)
+ *
+ * @author David G.
+ */
+public class SAMLBearerRequest extends TokenRequest {
+
+    static Map<String, Object> secret = DecryptUtil.decryptMap((Map<String, Object>) Config.getInstance().getJsonMapConfig(Http2Client.CONFIG_SECRET));
+    // x-www-urlencoded keys / values sent to OAuth server for SAML grant flow
+    static final String CLIENT_ASSERTION_TYPE_KEY = "client_assertion_type";
+    static final String CLIENT_ASSERTION_TYPE_VALUE = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
+    static final String CLIENT_ASSERTION_KEY = "client_assertion"; // value is JWT
+    static final String ASSERTION_KEY = "assertion"; // value is SAML token
+    static final String GRANT_TYPE_KEY="grant_type";
+    static final String GRANT_TYPE_VALUE="urn:ietf:params:oauth:grant-type:saml2-bearer";
+
+    private String samlAssertion = "";
+    private String jwtClientAssertion = "";
+
+
+    public SAMLBearerRequest(String samlAssertion , String jwtClientAssertion) {
+
+        setGrantType(SAML_BEARER);
+        this.samlAssertion = samlAssertion ;
+        this.jwtClientAssertion = jwtClientAssertion ;
+
+        Map<String, Object> clientConfig = Config.getInstance().getJsonMapConfig(Http2Client.CONFIG_NAME);
+        // client_secret is in secret.yml instead of client.yml
+        if(clientConfig != null) {
+            Map<String, Object> oauthConfig = (Map<String, Object>)clientConfig.get(OAUTH);
+            if(oauthConfig != null) {
+                Map<String, Object> tokenConfig = (Map<String, Object>)oauthConfig.get(TOKEN);
+                if(tokenConfig != null) {
+                    setServerUrl((String)tokenConfig.get(SERVER_URL));
+                    Object object = tokenConfig.get(ENABLE_HTTP2);
+                    setEnableHttp2(object != null && (Boolean) object);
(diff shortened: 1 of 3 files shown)


stevehu:
@dguncb I have reviewed the code technically and it looks good. No impact on the existing functionality. However, I am not very familiar with this new grant type and it looks like you are sharing the token service config section from the client.yml file. So please ensure that the config info is enough and the interaction with your OAuth 2.0 provider is tested. Let's wait for @miklish or @NicholasAzar to have another review. Thanks.

dguncb:
Yes, it is sharing much of the same configuration information from client.yml, such as URI, servername and port. Only the client_id and client_secret may not be needed and those 2 lines could be commented out. I have tested locally to see the request going to the server is the same as in the sample  provided by the Layer 7 specialist. On Monday we will be testing against the actual oauth server.

stevehu:
Thanks for the detailed explanation. Are you planning to add a middleware handler into the light-router? If yes, will you be able to make it into the planned release next weekend?

dguncb:
HI Steve,

Yes, just sent you the pull request for the middleware handler in the light router repo. Yes, we should be able to make it to the next release.

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
```diff
@@ -0,0 +1,84 @@
...
+    private String jwtClientAssertion = "";
+
+
+    public SAMLBearerRequest(String samlAssertion , String jwtClientAssertion) {
+
+        setGrantType(SAML_BEARER);
+        this.samlAssertion = samlAssertion ;
+        this.jwtClientAssertion = jwtClientAssertion ;
+
+        Map<String, Object> clientConfig = Config.getInstance().getJsonMapConfig(Http2Client.CONFIG_NAME);
```
NicholasAzar: Should not be parsing the config on each request. Can we move this out and make static?

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
```diff
@@ -0,0 +1,84 @@
...
+        // client_secret is in secret.yml instead of client.yml
+        if(clientConfig != null) {
+            Map<String, Object> oauthConfig = (Map<String, Object>)clientConfig.get(OAUTH);
+            if(oauthConfig != null) {
+                Map<String, Object> tokenConfig = (Map<String, Object>)oauthConfig.get(TOKEN);
+                if(tokenConfig != null) {
```
NicholasAzar: Can we add log statements if these are found to be null?

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
```diff
@@ -0,0 +1,84 @@
...
+    // x-www-urlencoded keys / values sent to OAuth server for SAML grant flow
+    static final String CLIENT_ASSERTION_TYPE_KEY = "client_assertion_type";
+    static final String CLIENT_ASSERTION_TYPE_VALUE = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
+    static final String CLIENT_ASSERTION_KEY = "client_assertion"; // value is JWT
+    static final String ASSERTION_KEY = "assertion"; // value is SAML token
+    static final String GRANT_TYPE_KEY="grant_type";
+    static final String GRANT_TYPE_VALUE="urn:ietf:params:oauth:grant-type:saml2-bearer";
+
+    private String samlAssertion = "";
+    private String jwtClientAssertion = "";
```
NicholasAzar: No need to initialize these to empty string

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * This holds values used to call the SAML Bearer grant flow from the OAuth Server.
+ * In this version client presents a SAML token and a JWT token (requirment for project)
+ *
+ * @author David G.
+ */
+public class SAMLBearerRequest extends TokenRequest {
+
+    static Map<String, Object> secret = DecryptUtil.decryptMap((Map<String, Object>) Config.getInstance().getJsonMapConfig(Http2Client.CONFIG_SECRET));
```
NicholasAzar: Redundant cast to `Map<String, Object>`

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
```diff
@@ -0,0 +1,84 @@
...
+import com.networknt.common.DecryptUtil;
+import com.networknt.common.SecretConstants;
+import com.networknt.config.Config;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This holds values used to call the SAML Bearer grant flow from the OAuth Server.
+ * In this version client presents a SAML token and a JWT token (requirment for project)
```
NicholasAzar: Can we remove the "requirement for project" comment here? This should be a solution against a pattern that can be used across many projects.

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
dguncb: Yes, this puzzled me a bit too. Steve, any reason was there a reason for not making the config values static in the existing ClientCredentialsRequest object ?

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
dguncb: I've found best practice is to inialize, you don't think so ?

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
dguncb: Yes, i put it in there because there could be confusion about why we have a JWT token in SAMLBearer grant type. This pattern is unlikely to be widely adopted outside CIBC in my opinion. And another weakness in this approach is that there is nothing tying the SAML to the JWT, so a potential attacker could mix SAMLs with JWTs. A improved approach may be for the SAML to be part of the JWT and the entire thing signed.

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
NicholasAzar: In this case where the actual values are assigned directly in the constructor, the initializing values will never be referenced

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
NicholasAzar: After speaking with the solution designer on this, this pattern could be used across many projects, and has application outside our team. It is a combination of publicly available RFC-7522 & RFC-7523.
For now we can assume that the security considerations have been thought through when designed, and if you find an actual case where a malicious subject or jwt can be swapped in or out, please raise it internally.
Until then, we can move forward with this design under the notion that is not a specific requirement for a project, and the comment can be updated to reflect this.

client/src/main/java/com/networknt/client/oauth/SAMLBearerRequest.java
stevehu: The clientConfig is not parsed for every request but retrieved from the cache so performance wise, it is the same. However, the best practice is to move it outside. Thanks.

NicholasAzar:
@dguncb are these changes going to be made soon?

dguncb:
yes, hopefully tonight.

PR #298, opened 2018-09-29
Develop

dguncb:
HI Steve, we have a request to make the hostname header configurable when calling the oauth server, rather than be hardcoded to locahost.

### DIFF
--- a/client/src/main/java/com/networknt/client/oauth/OauthHelper.java
+++ b/client/src/main/java/com/networknt/client/oauth/OauthHelper.java
@@ -136,8 +136,9 @@ public class OauthHelper {
                 public void run()  {
                     final ClientRequest request = new ClientRequest().setMethod(Methods.POST).setPath(tokenRequest.getUri());
                     request.getRequestHeaders().put(Headers.HOST, "localhost");
-                    request.getRequestHeaders().put(Headers.TRANSFER_ENCODING, "chunked");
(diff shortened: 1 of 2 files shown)


miklish:
David. I met with Zoran Roozebeh and Henry and it was decided to remove this requirement (for host header to match SAML Audience value). Please review email I sent to you Friday afternoon.

No changes should be made to existing SAML grant flow handler.

dguncb:
Hi Mike, yes i know but didn't you say we need it until Zoran makes his update in 2 weeks ?

miklish:
That's right. So I don't see any need to merge these changes into the foundation given they have a 2 week lifespan

dguncb:
ok..up to you and Nick..if you want to build off of the fork for the temporary solution.

miklish:
If you don't plan to use my code that pulls Audience directly from SAML, then check out the Header handler.

https://doc.networknt.com/concern/header/#readout

dguncb:
yeah this is very applicable to us. We'll use it next release of the SAML handler, there are a few updates to make.

miklish:
Hi Steve. We should reject this pull request. This code new code is only needed until Zoran returns on Oct 15 and updates L7 such that we no longer need to configure the Host header.

The plan is this: Until Oct 15 we will build images off of David's fork (for testing with L7's current policy). However, after Oct 15 (when Zoran updates the security policy for L7) we can go back to using tag/1.5.19 (since the updated policy will no longer check Host header).

PR #327, opened 2018-11-22
Fix H2DataSource issue with data-source module and refactor, add factory construct arguments feature to SingletonServiceFactory

miklish:
#### 1. Refactor data-source module
The data-source module is refactored to use delegation as the means for code reuse instead of inheritance. As part of this refactoring, `com.networknt.db.GenericDataSource` becomes an interface instead of a concrete base class. Reusable code is placed in `DataSourceHelper`.

#### 2. Fix case where H2DataSource params loaded for non-H2 DB's if no constructor arg provided 
Fix issue with `com.networknt.db.XDataSource` classes, whereby instantiating any such class with its default constructor results in the class trying to load H2DataDataSource parameters from data-source.yml. E.g.: instantiating `MySqlDataSource` will see `MySqlDataSource` attempt to use H2DataSource parameters from `data-source.yml` (instead of MySqlDataSource parameters).

#### 3. Add DI ability to provide parameters to factory constructor

These changes are in SingletonServiceFactory (e.g.: Light's Dependency Injection system).

##### Existing functionality
```yaml
singletons:
- javax.sql.DataSource: com.networknt.db.MysqlDataSource::getDataSource
```
In the example service.yml above, one can use a factory method of the constructed object to generate a class that implements the interface. Currently, one cannot provide arguments to the factory constructor.

##### New functionality

```yaml
singletons:
- javax.sql.DataSource:
  - com.networknt.db.MysqlDataSource::getDataSource:
    - java.lang.String: "MysqlDataSource"
```
The new syntax introduced (see above) allows one to specify parameters to the factory's constructor.

**_Note_**: The syntax shown under **Existing Functionality** above still works as it always has--no existing functionality of SingletonServiceFactory has been removed via this update.


data-source/src/main/java/com/networknt/db/H2DataSource.java
```diff
@@ -1,35 +1,25 @@
...
-    public String getDbPassKey() {
-        return H2_PASSWORD;
-    }
+    private DataSourceHelper _dsh;
 
-    public H2DataSource(String dsName) {
-        super(dsName);
-    }
+    public H2DataSource() { _dsh = new DataSourceHelper(DS_NAME, DB_PASS_KEY); }
+    public H2DataSource(String dsName) { _dsh = new DataSourceHelper(dsName != null ? dsName : DS_NAME, DB_PASS_KEY); }
```
NicholasAzar: Is there a reason to collapse everything? Makes this a bit hard to read imo..

service/src/main/java/com/networknt/service/SingletonServiceFactory.java
```diff
@@ -77,8 +77,14 @@ private static Object handleSingleImpl(List<String> interfaceClasses, List<Objec
         List<Object> items = new ArrayList<>();
         if (it.hasNext()) {
             Map.Entry<String, Map<String, Object>> pair = (Map.Entry) it.next();
+
+            // check whether intent is to generate implementation from factory method
             String key = pair.getKey();
-            Class implClass = Class.forName(key);
+            String initClassName = key.contains("::") ? key.substring(0, key.indexOf("::")) : key;
+            String initMethodName = key.contains("::") ? key.substring(key.indexOf("::") + 2) : null;
```
NicholasAzar: Can the "::" be moved to a constant, and the + 2 be referenced as the length of the constant?

service/src/main/java/com/networknt/service/SingletonServiceFactory.java
```diff
@@ -109,6 +115,13 @@ private static Object handleSingleImpl(List<String> interfaceClasses, List<Objec
                 }
             } else if(mapOrList instanceof List){
                 obj = ServiceUtil.constructByParameterizedConstructor(implClass, (List)mapOrList);
+
+                // use instance factory method to instantiate object if one was specified
+                if(initMethodName != null) {
+                    Method method = obj.getClass().getMethod(initMethodName);
+                    obj = method.invoke(obj);
```
NicholasAzar: Will the error message that results of the provided method name not belonging to the class be understandable?

service/src/test/java/com/networknt/service/N.java
```diff
@@ -0,0 +1,5 @@
+package com.networknt.service;
+
+public interface N {
```
NicholasAzar: What are these class names? Can they be changed to make their intention at least a bit clear?

service/src/test/java/com/networknt/service/SingletonServiceFactoryTest.java
```diff
@@ -134,5 +134,18 @@ public void testInitializerInterfaceWithBuilder() {
...
+        Assert.assertNotNull(n);
+        Assert.assertTrue(n.getVal().equals("default"));
+    }
+
+    @Test
+    public void testFactoryConstructorWithArguments() {
+        O o = SingletonServiceFactory.getBean(O.class);
+        Assert.assertNotNull(o);
+        Assert.assertTrue(o.getVal().equals("newVal"));
+    }
```
NicholasAzar: Could any negative test cases be added?

miklish:
Will do (though I do wish we used a prefix for (non static, non public) member variables -- e.g.: Google uses the prefix m in AOSP). As for condensed variable names, I assume you mean some of the constants in the XDataSource classes in the  data-source module? If so, will do. Will also create PR for docs update.

chenyan71:
The fix looks good. 

I am using this module for one of my project. And as the issue described, if we pass the data-source as below:

singletons:
- javax.sql.DataSource: com.networknt.db.MysqlDataSource::getDataSource

It will only call default constructor and then use the default DS_NAME = "H2DataSource"
I created work around:

- com.networknt.db.GenericDataSource:
  - com.networknt.db.OracleDataSource:
    - java.lang.String: OracleDataSource

And then pass the GenericDataSource for DAO class, get datasource from DAO constructor by call GenericDataSource.getDataSource().

But we need make change on test case and test config file which I didn't see in the PR:

https://github.com/networknt/light-4j/blob/develop/data-source/src/test/resources/config/service.yml

stevehu:
@chenyan71 Could you please help to write the document on how to use it? We can merge it once the document is done. Thanks.

stevehu:
This is not needed at the moment. Close it.

Issue #326, opened 2018-11-13
Attachment Handler


thirtysixmm:
A middle ware handler to parse the form-data based on the header and put the content into an attachment. It is an optional handler that you can wire into your handler chain for a specific endpoint.

NicholasAzar:
Will take a look

NicholasAzar:
This is being addressed in: https://github.com/networknt/light-4j/pull/329

stevehu:
@thirtysixmm It is scheduled to be released this weekend. 

https://trello.com/b/189msq9S/release-schedule

PR #329, opened 2018-12-06
#326 created a form handler

jiachen1120:
@NicholasAzar 
Goal
    - Create a middleware handler for parsing form-data

Background
    - Form-Data
       Multipart/Form-Data is a format used for transferring form data (information typed into a web page). It is a simplified form of MIME.
       It is one of the formats that may constitute the body of an HTTP "POST" message. It serves the same purpose as Form URL encoding, but is more efficient when transferring large amounts of binary data. Web browsers may require it for file uploads.

Overview/Requirement
    - Build a form-data handler which can be used to paring form-data and attach result to exchange
Design
    - Check the content-type within the request header (x-www-form-urlencoded or multipart/form-data) 
    - Build a FormDataPaser by using FormDataPaserFactory(only x-ww-form-urlencoded be used to generate the parser, so I convert the content-type to match it)
    - Start parse and attach it to the exchange

Testing
    - test get method
    - test post method with invalid form(fail)
       reason: Parser can parse everything. Therefore, although the request is invalid, there still a 200, is it necessary to write a data-form validation?
    - test post method with valid form
    - test post method without contant-type header
    - add to handler chain of an example app and test in postman 

Alternate
    - Is it better to combine the bodyhandler and formhandler together to let the handler to match the right parser automatically? therefore, the body module can run without consider what kind content-type go through.


NicholasAzar:
> Alternate
> - Is it better to combine the bodyhandler and formhandler together to let the handler to match the right parser automatically? therefore, the body module can run without consider what kind content-type go through.

Yes! I think it's a great idea to enhance the BodyHandler to parse based on content type. FYI @stevehu

NicholasAzar:
[Parsing form data into an attachment](https://trello.com/c/BYyBopis/39-parsing-form-data-into-an-attachment)

body/src/main/java/com/networknt/body/BodyHandler.java
```diff
@@ -106,6 +115,30 @@ public void handleRequest(final HttpServerExchange exchange) throws Exception {
...
+            if (contentType.startsWith("multipart/form-data")) {
+                exchange.getRequestHeaders().put(Headers.CONTENT_TYPE, "application/x-www-form-urlencoded");
+            }
+            try {
+                FormParserFactory formParserFactory = FormParserFactory.builder().build();
+                FormDataParser parser = formParserFactory.createParser(exchange);
+                if (parser != null) {
+                    FormData data = parser.parseBlocking();
+                    exchange.putAttachment(REQUEST_BODY, data);
+                }
```
NicholasAzar: Would it be possible to convert the FormData into a Map so the endpoint's receive a consistent data type?

NicholasAzar:
Docks pr: https://github.com/networknt/light-doc/pull/28

NicholasAzar:
Docs pr: https://github.com/networknt/light-doc/pull/28

Issue #332, opened 2018-12-11
In consul client the hard coded wait value changed through configuration. 


sreenicibc:
In consul client the hard coded wait value changed through configuration.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [In consul client the hard coded wait value changed through configuration](https://trello.com/c/RlGahplv/45-in-consul-client-the-hard-coded-wait-value-changed-through-configuration)

PR #333, opened 2018-12-11
Issue #332, (API 6) in consul client the wait hard coded value changed through configuration 

sreenicibc:
Issue #332, (API 6) in consul client the wait hard coded value changed through configuration.

### DIFF
--- a/consul/src/main/java/com/networknt/consul/ConsulConfig.java
+++ b/consul/src/main/java/com/networknt/consul/ConsulConfig.java
@@ -8,6 +8,8 @@ public class ConsulConfig {
     boolean tcpCheck;
     boolean httpCheck;
     boolean ttlCheck;
+    String wait;
+
 
     public String getConsulUrl() {
         return consulUrl;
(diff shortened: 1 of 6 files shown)


stevehu:
I was considered to make it configurable in the beginning but hard-coded instead. This is the value to controls [blocking query](https://www.consul.io/api/index.html#blocking-queries) timeout for consul. It has a hard limit of 10 minutes and default to 5 minutes. I have decided to use 10 minutes as default as it is the maximum we can use. I am afraid people will put something like 900s and I don't know what would be the behavior of Consul in that case. If we want to make it configurable, I think we need to make more info available in the consul.yml file about this property to ensure that users won't misconfigure this field. Do we have other reasons we want to change this value? For TSO project? Thanks.

sreenicibc:
Added description about that property in consul.yml file.

NicholasAzar:
@stevehu are those comments ok or did you have anything else in mind? I can't remember which projects have asked for this change, but its come up more then once..

stevehu:
With the comment in the config file, I think it will be less risky to make any mistakes. I am OK with the change. Thanks.

Issue #280, opened 2018-09-06
need a way to serialize json map object to String for logging with mask


stevehu:
In most of the handlers, we are dealing with Map object for the input JSON. For logging purpose, we need to serialize it into a string and mask the sensitive data like password etc. I don't want to use Jackson writeValueAsString for two reasons. 

1. We need to handle the exception
2. There is no way we can mask sensitive values. 

Create two static methods in Mask? One deals with a list and another deals with a map.

stevehu:
I agree. That is why I was thinking to extend Mask module and utilize its config file.

NicholasAzar:
@BalloonWen is this done?

BalloonWen:
@NicholasAzar This was done and tested at that time but after the change of JsonPathStartupHookProvider, it may have some affection. I haven't tested after that.

NicholasAzar:
Ok thanks, i'm thinking we should track any impact from other changes in a new issue.

Issue #23, opened 2017-02-07
refactor the dump handler to support response body logging


stevehu:
Need to upgrade to the latest Undertow version which supports response body dump.

BalloonWen:
Newer undertow did something to support this feature, but it has some defects.
I did implementation based on a similar concept.

PR #345, opened 2019-01-02
refactored dump handler, add mask feature to it

BalloonWen:
-refactored dump logger
-fixed defects when combining with BodyHandler, this handler should be after BodyHandler, because after reading from InputStream, the stream will be unavailable, the body won't get the attachment anymore
-dump logger now can log on a different level (ERROR, WARN, DEBUG, INFO) based on config.
-add mask feature when dumping request and response. this should be configured in mask.yml
related issue: #23 #280


BalloonWen:
> @BalloonWen Could you add class & method level comments for a lot of the dumper and mask classes describing what they're used for and how to use them?

comments added, please advise if you have any concerns.

NicholasAzar:
Docs pr for this change: https://github.com/networknt/light-doc/pull/35

Issue #321, opened 2018-11-05
Environment variable references in the light-4j yaml config files


pragmaticway:
This is enhancement request to create a "preprocessor" for yaml configuration files to be able to inject environment variables.

Proposed implementation specification:
--------------------------------------------
You can use environment variable references in the config file to set values that need to be configurable during deployment. To do this, use:

`${VAR}`

Where VAR is the name of the environment variable.

Each variable reference is replaced at startup by the value of the environment variable. The replacement is case-sensitive and occurs before the YAML file is parsed. References to undefined variables are replaced by empty strings unless you specify a default value or custom error text.

To specify a default value, use:

`${VAR:default_value}`
Where default_value is the value to use if the environment variable is undefined.

To specify custom error text, use:

`${VAR:?error_text}`

Where error_text is custom text that will be prepended to the error message if the environment variable cannot be expanded.

If you need to use a literal ${ in your configuration file then you can write `$${` to escape the expansion.

NicholasAzar:
Will take a look, this is a very interesting proposition.

ddobrin:
I would like to add a suggestion to this issue, within an existing module. 

I am discussing the following approach with my current client, and they are preparing a file with the environment variables they wish to override as an API moves across test phases. 

- the Config module loads the configuration for each individual module
- at the time of the loading of the config, an environment variable resolution process can kick in (as per @pragmaticway )
- the APIs are loading the configuration from a single location, the externalized folder (Config.java L.279). The file with the customizations can either be loaded from a set file (say env_vars.yml) within the same folder or the Config module can be expanded to allow different folders for config loading, similar to the approach we have used in (https://github.com/networknt/light-bot). 
- the customized file can be served via config-server
-  if the externalized file is present, {vars} can be resolved, with defaults or not, etc

@stevehu @NicholasAzar @pragmaticway : what do you think?

miklish:
Much interest, research, and development time has been spent on the configuration issue recently. David is looking into possible config server enhancements and I have been developing a similar config-value injection tool myself. We should have a design session to ensure what results is a single coherent, integrated approach and not a loose collection of tools that may or may not perform the same functions

ddobrin:
@miklish : how do you propose that we go about this session?
I need to put a process in place in the next 1-2 weeks for it, for 2 separate initiatives.

There have been a number of configuration mishaps at a client site and this, in addition to config server and some other aspects not mentioned here will alleviate them in good measure.

stevehu:
[...]

PR #348, opened 2019-01-07
Config Injection

jiachen1120:
@stevehu @NicholasAzar Implements config injection from both environment and file called `values.yaml`. A RFC associates with this PR also could be found from another PR.


jiachen1120:
https://github.com/networknt/light-4j/issues/321 This request was closed because another request which is more powerful and supporting List and Map injection was pulled. I will add some comment into code later. you can check it over there: https://github.com/networknt/light-4j/pull/349

PR #349, opened 2019-01-08
#321 Enhanced config injection to support injecting List and Map

jiachen1120:
@NicholasAzar @stevehu @ddobrin 
I implements this enhancement as the following steps:
1. parse config into Map
2. search the ${} recursively
3. injection

The MD can be found here: https://github.com/jiachen1120/light-rfcs/blob/develop/light-4j/0006-config-injection-guidance.md


stevehu:
@jiachen1120 Thanks for your help.

PR #361, opened 2019-01-22
Fixes #360 If handler path not implemented/wrong path , it will throw the 404 error code instead of 400.

### LINKED ISSUE #360: If handler path not implemented/wrong path , it will throw the 404 error code instead of 400.
API 30 -If handler path not implemented or wrong path , it will throw the 404 error code instead of 400.

sreenicibc:
Fixes #360 If handler path not implemented/wrong path , it will throw the 404 error code instead of 400.

### DIFF
--- a/handler/src/main/java/com/networknt/handler/OrchestrationHandler.java
+++ b/handler/src/main/java/com/networknt/handler/OrchestrationHandler.java
@@ -7,7 +7,7 @@ import io.undertow.server.HttpServerExchange;
  */
 public class OrchestrationHandler implements LightHttpHandler {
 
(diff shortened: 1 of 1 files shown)


NicholasAzar:
@stevehu I believe this is a regression from when we moved away from having the Openapi validator be able to produce the path not found responses. It use to return this status code ERR10007 from OpenApiHandler.java.
Are you ok with going back to 404?

stevehu:
Before the handler module is introduced, this error is captured by the validator of each framework. Now it is captured in the handler module. In my opinion, both 400 and 404 are OK as long as the error message is clear.

Issue #362, opened 2019-01-23
Add a new constant string in http header for service url


jsu216:
This issue is for the Jira ticket API-36
It allows the client of the light-router to specify the service url where the requests should be forwarded to.
Another issue, https://github.com/networknt/light-router/issues/38 is created in light-router to support this feature.

stevehu:
@jsu216 @NicholasAzar As per our discussion, this will be addressed internally. Closing the issue.

PR #363, opened 2019-01-23
Add a new constant in http header for service_url.

jsu216:
This is for the issue #362  and Jira ticket: API-36.


NicholasAzar:
Closed, per discussion with @zabooma & @jsu216 , will be addressed elsewhere.

PR #364, opened 2019-01-23
-added an error message in status.yml for validating response content

BalloonWen:
related issue: https://github.com/networknt/light-codegen/issues/133


stevehu:
@BalloonWen There are different sections in the status.yml file and we pre-allocate ranges for each type of error codes. This error should be in the validation error section. Thanks.

BalloonWen:
@stevehu now it's moved where it should be.

Issue #365, opened 2019-01-23
Standardize repo line endings on LF


NicholasAzar:
Currently this repo is mixed between LF (primarily) and CRLF (in a few scattered files).
We are seeing this issue manifest in PRs that are submitted with LF (with only a few lines changed) that get reported as the entire file being modified and makes it difficult to identify what's being modified.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Standardize repo line endings on LF](https://trello.com/c/Ea8PYb4D/72-standardize-repo-line-endings-on-lf)

stevehu:
Merged #366.

PR #366, opened 2019-01-23
Fix/line endings LF

NicholasAzar:
Change to address #365


stevehu:
@NicholasAzar We need to exclude crt files :)

NicholasAzar:
@stevehu done, even though from what I saw all *.crt's where base64.. don't know if there's another reason they need to have their crlf's maintained.

stevehu:
@NicholasAzar Thanks. I don't know if there is any impact at all but it is safe to keep the way it is.

PR #367, opened 2019-01-24
Feature/single config for all modules

jiachen1120:
@NicholasAzar @stevehu 
Implemented single config for all modules, look forward to your review.
The RFC related to the PR: https://github.com/jiachen1120/light-rfcs/blob/develop/light-4j/0008-single-config-for-all-modules.md


NicholasAzar:
Closing this since it doesn't seem like we'll be moving with this solution, however please keep the branch available.

PR #372, opened 2019-01-30
Fixing mutual tls on the server side

NicholasAzar:
@stevehu been looking into getting mutual tls on the server end working and there seem to be some issues.
I first noticed that this check seems to be inverted, but even after this fix it doesn't seem to be restricting access to only a client who can present the cert in the server truststore.

Have you ever had this working on your end?


NicholasAzar:
@stevehu I tested @dz-1 change locally and it worked great, please assign another reviewer as you see fit

stevehu:
[Fixing mutual tls on the server side](https://trello.com/c/0HSbwHvj/71-fixing-mutual-tls-on-the-server-side)

PR #373, opened 2019-01-30
Allowed exclusion list of files which should not check/allow values injection

jiachen1120:
This PR is related to https://github.com/networknt/light-4j/issues/368

Just simple changes by loading the exclusions.yml in the ConfigInjection class and provide a method to check whether the current filename is listed in the exclusions.yml. Then invoke it to check before using values injection. 

Look forward to your opinion! @NicholasAzar @stevehu @ddobrin

### DIFF
--- a/config/src/main/java/com/networknt/config/Config.java
+++ b/config/src/main/java/com/networknt/config/Config.java
@@ -205,8 +205,8 @@ public abstract class Config {
             String ymlFilename = configName + CONFIG_EXT_YML;
             try (InputStream inStream = getConfigStream(ymlFilename)) {
                 if (inStream != null) {
-                    // The values.yaml should be processed by centralized management, since it would cause a dead loop
-                    if (configName.equals("values")) {
+                    // The config file specified in the exclusions.yml shouldn't be injected
+                    if (ConfigInjection.isExclusionConfigFile(ymlFilename)) {
                         config = yaml.loadAs(inStream, clazz);
                     } else {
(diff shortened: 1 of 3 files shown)


stevehu:
It was my idea to use exclusions.yml for this purpose but after talking to @NicholasAzar and @jiachen1120 , I realized that the file name is too specific. The idea is to change the name to config.yml with a section called exclusions with a list of filenames. With this proposal, we follow the same convention that name the config file as the module name and leave the room in the future if we want to have more variables to further customize the config module. Additionally, we need to bypass the environment variable injection for this file when loading it. @NicholasAzar @jiachen1120 please correct me if I miss some points. @ddobrin You have approved the original design with filename exclusions.yml. Are you OK with the filename change? Thanks.

ddobrin:
I have proposed this list to be in the config.yml to start with, I was overruled, therefore I certainly agree with using config.yml. 
It makes sense

jiachen1120:
In terms of the format. Can we just list the file names in the list instead of listing the file names.extension? Because the configuration module can handle the suffix problem.

So, it will look like:
exclusionConfigFileList:
	- openapi
	- values
	- status
	
In other words, is it necessary to distinguish files based on suffixes? Look forward to your opinions.

stevehu:
@jiachen1120 I think it is a good idea to just list the file names regardless extensions.

NicholasAzar:
@stevehu you may want to take another look before merging, I made some changes to reduce code duplication in the Config class.

PR #379, opened 2019-02-07
fixes #378 upgrade jsonpath to 2.4.0 and exclude json-smart

### LINKED ISSUE #378: upgrade jsonpath to 2.4.0 and exclude json-smart
During the testing on the develop branch, I realized that we are including json-smart from JsonPath as it is the default JSON parser provider. The JsonPath is only used by the mask module but we don't want to load two json parsers in the same application. 

To replace the json-smart with Jackson: 

* change the parent pom.xml to exclude the json-smart from jsonpath dependency.
* update Mask.java to remove the json-smart specific class.
* update MaskTest.java to add a BeforeClass to config the JsonPath with Jackson provider.
* add JsonPathStartupHookProvider in server module.
* update the document site to indicate that the user needs to include a StartupHookProvider to config JsonPath if mask module is used.

stevehu:


stevehu:
@ddobrin I think it is a good idea to document all the handlers and in which case these handlers will be used. I am just not sure if this should be defined in each individual frameworks or in the light-codegen section in the document site. Or the other options is to write the comments in the handler.yml file in details.

PR #380, opened 2019-02-07
Feat/#41 exception handling

BalloonWen:
related RFC document: https://github.com/networknt/light-rfcs/commit/f440c47e7b2931995ec7263f6496063106d76416
related issue: https://github.com/networknt/light-router/issues/41
**before change**: in client module whenever there's an error, throw exceptions, then ExceptionHandler will catch it and respond to the client (lost detail information)
**after change**: return a com.networknt.monad.Result, the user can handle based on the "Result" is success or failure. "Result" will keep failure information if fails.

### DIFF
--- a/client/src/main/java/com/networknt/client/oauth/Jwt.java
+++ b/client/src/main/java/com/networknt/client/oauth/Jwt.java
@@ -1,0 +1,101 @@
+package com.networknt.client.oauth;
+
+import com.networknt.config.Config;
+
+import java.util.Map;
+
+public class Jwt {
+    private String jwt;    // the cached jwt token for client credentials grant type
+    private long expire;   // jwt expire time in millisecond so that we don't need to parse the jwt.
+    private volatile boolean renewing = false;
+    private volatile long expiredRetryTimeout;
+    private volatile long earlyRetryTimeout;
+
+    private static long tokenRenewBeforeExpired;
+    private static long expiredRefreshRetryDelay;
+    private static long earlyRefreshRetryDelay;
+
+    static final String OAUTH = "oauth";
+    static final String TOKEN = "token";
+    static final String TOKEN_RENEW_BEFORE_EXPIRED = "tokenRenewBeforeExpired";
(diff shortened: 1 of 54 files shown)


BalloonWen:
now whenever needs to pass auth server's response to the client, now it will escape based on the type that the server returns. @NicholasAzar 
https://github.com/networknt/light-4j/pull/380/commits/4b354196cd5b77cf2315e3343088f6e0d776943c

NicholasAzar:
@BalloonWen 
{"statusCode":401,"code":"ERR10052","message":"GET_TOKEN_ERROR","description":"Cannot get valid token: "{\\r\\n  \\"error\\":\\"invalid_client\\",\\r\\n  \\"error_description\\":\\"The given client credentials were not valid\\"\\r\\n}".","severity":"ERROR"}

Json format error is: 
"Expected ',' instead of '{'"

NicholasAzar:
Looks better! : 
{
    "statusCode": 401,
    "code": "ERR10052",
    "message": "GET_TOKEN_ERROR",
    "description": "Cannot get valid token: {\r\n  \"error\":\"invalid_client\",\r\n  \"error_description\":\"The given client credentials were not valid\"\r\n}.",
    "severity": "ERROR"
}

client/src/main/java/com/networknt/client/oauth/Jwt.java
```diff
@@ -0,0 +1,101 @@
+package com.networknt.client.oauth;
+
+import com.networknt.config.Config;
+
+import java.util.Map;
+
+public class Jwt {
```
NicholasAzar: Class level comments

http-string/src/main/java/com/networknt/httpstring/ContentType.java
```diff
@@ -0,0 +1,26 @@
+package com.networknt.httpstring;
+
+public enum ContentType {
```
NicholasAzar: Class level comments pretty please

Issue #309, opened 2018-10-16
Read configurations from arbitrary directories


ddobrin:
The Config class offers very convenient methods to load config files in yml or json formats.

The are predicated on being available in the "light-4j-config-dir", which is specified at startup as usually with a "-D".

The Config.loadMapConfig(String fileName) could be overloaded with a directory location as:
Config.loadMapConfig(String fileName, String directoryName).

It would be useful for : 
- tools as the networknt:light-bot project for loading config files from multiple locations
- specifying configurations in multiple folders, in addition to the standard /src/main/resources/config

NicholasAzar:
@ddobrin @stevehu Is there an rfc for this?

ddobrin:
https://github.com/networknt/light-rfcs/pull/11

ddobrin:
RFC has been split into 3 smaller RFCs, by problem area.
It can be found here:
https://github.com/networknt/light-rfcs/pull/15

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Read configurations from arbitrary directories](https://trello.com/c/5J1gLc4Z/82-read-configurations-from-arbitrary-directories)

PR #386, opened 2019-02-11
Feat/#309 read config from arbitrary directories

jiachen1120:
Related issue: https://github.com/networknt/light-4j/issues/309
Related RFC: https://github.com/networknt/light-rfcs/blob/master/light-4j/0012-flexible-config-directory.md

Three new overloaded methods provided, in this PR, the path can only be the path relative to light-4j-config-dir. 
public abstract Map<String, Object> getJsonMapConfig(String configName, String path);
public abstract Map<String, Object> getJsonMapConfigNoCache(String configName, String path);
public abstract Object getJsonObjectConfig(String configName, Class clazz, String path);

Extended ConfigDefaultTest to included the test for these methods. The ConfigDefaultTest could not work properly before if build the whole light-4j project entirely. Since the config is a singleton and shared by all classes, the instance of config would already exist before running ConfigDefaultTest. If we set the light-4j-config-dir directly, it cannot be injected into config instance. So the config in ConfigDefaultTest would still be the classpath one. So now I use reflection to inject field in the test.

Look forward to your review!


config/src/main/java/com/networknt/config/ConfigInjection.java
```diff
@@ -160,4 +169,29 @@ public void setKey(String key) {
...
     }
+
+    // Method used to cast string into int, double or boolean
+    private static Object typeCast(String str) {
+        if (str == null || str.equals("")) {
+            return null;
+        }
+        if (isBoolean(str)) {
+            return str.equals("true") || str.equals("True") || str.equals("TRUE") ? true : false;
+        }
```
NicholasAzar: Can be simplified to just `return str.equals("true") || str.equals("True") || str.equals("TRUE");` right?

miklish: or
return str.toLowerCase().equals("true");

jiachen1120: Yes, I will simplify it. @miklish The reason why I didn't use toLowerCase() is to uniform with yaml rules, ie "tRue", "trUe", etc. should not be recognized as boolean. Do you think it makes sense?

miklish: It simply depends on how strict you want your parsing to be in this instance.

NicholasAzar: @jiachen1120 Agreed, we shouldn't use toLowerCase.
Actually, if you want to be even closer to the [yaml boolean spec](https://yaml.org/type/bool.html), we would use these:
```
 y|Y|yes|Yes|YES|n|N|no|No|NO
|true|True|TRUE|false|False|FALSE
|on|On|ON|off|Off|OFF
```

jiachen1120: Sure, I will improve it to be more strict.

jiachen1120: And I will modify the light-doc as well.

PR #401, opened 2019-02-18
body: update and add more unittest

lanphan:
update body module:
1. update / add more unittests to make it more details
2. add missing license info
3. update logback-test.xml

### DIFF
--- a/body/src/main/java/com/networknt/body/BodyHandler.java
+++ b/body/src/main/java/com/networknt/body/BodyHandler.java
@@ -36,8 +36,8 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


body/src/main/java/com/networknt/body/BodyHandler.java
```diff
@@ -69,7 +69,7 @@
     private volatile HttpHandler next;
 
     public BodyHandler() {
-
+        if(logger.isInfoEnabled()) logger.info("BodyHandler is loaded.");
```
bkuberek: any reason why this should be `INFO` as opposed to `DEBUG` ?

stevehu: A handler is a singleton and only loaded once, that is why we make it as info in case project team want to double check what kind of handlers are loaded as they are all plugins. If you see there is no issue, please approve it so that we can merge it. We need two reviewers to ensure the quality of the code. Thanks.

PR #402, opened 2019-02-19
add missing license info, remove unused logback setting

lanphan:
1. add missing license info
Please review carefully io.dropwizards package in metrics/ module, there are still some files I don't know license to add

This is the first step to prepare adding apache rat plugin to auto check license header for light-4j project.

2. remove unused logback setting


stevehu:
> Why are we removing the audit logger? I am assuming because it only makes sense to configure that in the application logger.

I think we only remove it from the logback-test.xml and there is no need to have audit logs as they are for module testing.

Issue #407, opened 2019-02-22
Error happen if there is not value in values.yml for services.yml injection


chenyan71:
Light-codegen (after 1.5.30) will generate service.yml as below:

```service.yml
### Singleton service factory configuration/IoC injection
singletons: ${service.singletons}

```

And in the values.yml, we can add value injection for the service.yml. But if there is not value need be set for services.yml. For example:

```values.yml
service.singletons:
# StartupHookProvider implementations, there are one to many and they are called in the same sequence defined.
# - com.networknt.server.StartupHookProvider:
  # If you are using mask module to remove sensitive info before logging, uncomment the following line.
  # - com.networknt.server.JsonPathStartupHookProvider
  # - com.networknt.server.Test1StartupHook
  # - com.networknt.server.Test2StartupHook
```
The system will get error when we start service:

Exception in thread "main" java.lang.ExceptionInInitializerError
	at com.networknt.server.Server.start(Server.java:143)
	at com.networknt.server.Server.init(Server.java:128)
	at com.networknt.server.Server.main(Server.java:115)
Caused by: com.networknt.config.ConfigException: "${service.singletons}" appears in config file cannot be expanded
	at com.networknt.config.ConfigInjection.getInjectValue(ConfigInjection.java:56)

But if we set some values in the values.yml, then it works fine:

```values.yml
service.singletons:
- com.networknt.balance.LoadBalance:
  - com.networknt.balance.RoundRobinLoadBalance
- com.networknt.cluster.Cluster:
  - com.networknt.cluster.LightCluster
```

stevehu:
@chenyan71 It looks like there is only a key without a value. We need to try something like an empty string to see if it works. @jiachen1120 is looking into it as well. Please have a chat with him to find the best solution. Thanks.

ddobrin:
It works without setting a value:

### Singleton service factory configuration/IoC injection
singletons:

I am using it as such in service.yml at my client.
The issue can arise if you leave maps/arrays generation on in the codegen, however this can easily be bypassed

jiachen1120:
Good morning! @chenyan71 @stevehu @ddobrin 

I tried returning an empty string and null for upper case provided by @chenyan71 . I found that an empty string would lead to an exception since the service module cannot parse the empty string. But `null` can pass normally.

The improved performance is as follows:

1. If can't get the value from either environment variables, default values, or values.yml. but values.yml contains the corresponding key, return `null`.

This improvement allows users to set null in values.yml. 
[...]

PR #409, opened 2019-02-25
fix/#407 the value in values.yml can be set to empty

jiachen1120:
Related issue: https://github.com/networknt/light-4j/issues/407


jiachen1120:
@ddobrin Sounds great! 👍 I already changed to what you suggested. Also, a test case is added to test null value scenario.

Issue #413, opened 2019-02-28
Integrate decryption module in all retrievals of config


dz-1:
Originated from API-49:
Investigation Required
Currently when config is retrieved it needs to be done so with explicit instruction to support decrypting values. This could lead to some modules not supporting encryption/decryption and could result in issues.

stevehu:
In the beginning the ideas is to group all the secrets into one file secret.yml and only enable decryption for that file. Apparently, it won't be enough and cause a lot of confusions for developers. If we are decrypt for every config file, then I think we should retire the secret.yml. What do you think?

dz-1:
Personally, I think putting secrets in one file is not bad and we can leave that file as is.
The fix of this issue gives developers more flexibilities in organizing files.

 @NicholasAzar What do you think?

NicholasAzar:
Moving those fields around could lead to a breaking change though right? Currently I can think of the following places where we have secrets external to secret.yml; metrics.yml, server.yml, basic-auth.yml.. looking at those I don't know if it's feasible to move everything secret to secret.yml.

stevehu:
We can still keep the secret.yml but make it deprecated and removed in 2.0.x later on. For an existing application, if they are using secret.yml, they must load it explicitly. The question is should we move those secrets back into individual config files.

ddobrin:
if the values are moved back into individual files, we need to think about this being documented very very clearly, as many of these "secret" values are set in the secret.yml + already set in the values.yml as overwrites.

Consumers would have to move them all back to their individual config files and, if you have a larger number of APIs, it has a ripple effect across all of them.

PR #414, opened 2019-02-28
Integrate decryption module in all retrievals of config - #413

dz-1:
With this change, all values loaded via yml parse are decrypted. In other words, the decrypting logic is transparent to config users. No explicit decrypting is needed anymore. Unnecessary references to the class `DecryptUtil` are cleaned up within light-4j projects.

Note: the `Decryptor` interface is moved from the `utility` project to the `decryptor` project in order  to avoid circular dependencies.
Also, the constant `FRAMEWORK_NAME` in `com.networknt.utility.Constants` is redefined in `com.networknt.decrypt.AESDecryptor` in order to reduce dependencies of the `decryptor` project.


stevehu:
@dz-1 Do you have any suggestion for document update so that users can leverage this feature without the limitation imposed before. Thanks.

dz-1:
@stevehu I can add a short section to `content/concern/config.md` of light-doc about this change. @jiachen1120 can add doc about his change later.

But we also need to let users know that the interface `com.networknt.utility.Decryptor` has been moved to `com.networknt.decrypt.Decryptor` somehow. Previous code could break if `com.networknt.utility.Decryptor` is used.

dz-1:
@stevehu document of this change has been added to [config.md](https://github.com/networknt/light-doc/blob/develop/content/concern/config.md)

stevehu:
@dz-1 Thanks a lot for the doc update. Given this might break the existing code, we will release it as 1.6.0 instead of 1.5.33. Thanks for the reminder. I will create a branch based on 1.5.32 for 1.5.x release branch and only bug fixes will down port there.

PR #416, opened 2019-03-02
Fixing binding to dynamic ports

### LINKED ISSUE #415: Dynamic Port binding not working as expected in light4j
Please find the bug details below.

Qualified Java Name: com.networknt.server.Server.bind(HttpHandler, int)

Runtime exception is thrown while binding the port if the port is not available. 
Expected behaviour is to loop through the port range and start with the one which is available. Thanks.

NicholasAzar:
Closes #415

### DIFF
--- a/server/src/main/java/com/networknt/server/Server.java
+++ b/server/src/main/java/com/networknt/server/Server.java
@@ -160,20 +160,11 @@ public class Server {
         }
 
         if (config.dynamicPort) {
-            boolean boundToPort = false;
             for (int i = config.minPort; i < config.maxPort; i++) {
-                try {
-                    boundToPort = bind(gracefulShutdownHandler, i);
-                    if (boundToPort) {
-                        break;
-                    }
-                } catch (RuntimeException ignore) {}
-            }
-            if (!boundToPort) {
-                String err = "No ports available to bind to. Tried: " + config.minPort + " to: " + (config.maxPort - 1);
-                System.out.println(err);
-                logger.error(err);
-                throw new RuntimeException(err);
+                boolean b = bind(gracefulShutdownHandler, i);
+                if (b) {
+                    break;
+                }
             }
         } else {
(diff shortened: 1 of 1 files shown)


NicholasAzar:
FYI @kumasath

stevehu:
Make a fix that reserve the previous behavior. When binding is failed, don't throw RuntimeException to stop the server but simply log one line message without the stack trace to indicate the binding to the port is failed, try the next port. Then return false to indicate that binding is failed so that the next port will be tried. In the second try/catch block for the consul registration, a RuntimeException is thrown and the server will be stopped. 

```
        } catch (Exception e) {
            System.out.println("Failed to bind to port " + port + ". Trying " + ++port);
            if (logger.isInfoEnabled())
                logger.info("Failed to bind to port " + port + ". Trying " + ++port);
            return false;
        }
```

stevehu:
[Fixing binding to dynamic ports](https://trello.com/c/Cds0wqnI/107-fixing-binding-to-dynamic-ports)

jiachen1120:
@NicholasAzar @stevehu Hey! I am very sorry that I didn't fix this bug in time.

I just did a manual test of develop branch and found the following problems:

If the user is using a static port and the port happens to be occupied, the user will get the following information:

Fail to bind to port 8453, trying 8454

And at this point the server does not stop, perhaps this will make the user mistakenly think that the server is always trying to bind the port. I think we should let the exception throw at the moment. 

Then, I test @NicholasAzar changes. due to the whole bind() is put into try catch block, the registration exception would be catched by outer try-catch and the user would got the port bind fail message when the registration fail.

In summary, I made some changes to make sure the following perfemance:
1. (dynamic port) keep trying until reach valid port. If registration fail, server stop after tried first port.
2. (static port) tring to bind port. If fail, throw exception. 

The PR: https://github.com/networknt/light-4j/pull/417

PR #417, opened 2019-03-02
Fix/#415 dynamic port binding

jiachen1120:
@NicholasAzar @stevehu Hey! I am sorry that I have not fulfilled my duties to complete the task you handed in. I just did a test and found that there may still be some areas to improve on this issue.

I just did a manual test of develop branch and found the following problems:

If the user is using a static port and the port happens to be occupied, the user will get the following information:

Fail to bind to port 8453, trying 8454

And at this point the server does not stop, perhaps this will make the user mistakenly think that the server is always trying to bind the port. I think we should let the exception throw at the moment. 

Then, I test @NicholasAzar changes. due to the whole bind() is put in to try catch block, the registration exception would be caught by outer try-catch and the user would get the port bind fail message when the registration fails.

In summary, I made some changes to make sure the following performance:
1. (dynamic port) keep trying until reach valid port. If registration fails, the server stop after tried the first port.
2. (static port) trying to bind port. If fail, throw the exception. 

Related topics:
https://github.com/networknt/light-4j/pull/416
https://github.com/networknt/light-4j/issues/415

### DIFF
--- a/config/src/main/java/com/networknt/config/Config.java
+++ b/config/src/main/java/com/networknt/config/Config.java
@@ -43,7 +43,6 @@ import java.util.concurrent.ConcurrentHashMap;
  */
 public abstract class Config {
     public static final String LIGHT_4J_CONFIG_DIR = "light-4j-config-dir";
-    public enum InternalizedConfigPath {DEFAULT, CLASSPATH}
 
     protected Config() {
     }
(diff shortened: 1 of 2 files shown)


NicholasAzar:
@jiachen1120 I think minPort should be ok to equal maxPort, just when it's strictly greater should there be an error

jiachen1120:
@NicholasAzar I also think it should be ok to equal in theory, but the previous performance is that the range of the port is: minPort <= port < maxPort. If let minPort equal to maxPort, no error message appear and no port bound successfully. So in order to make the previous performance unchanged I make such kind of change. @stevehu Or, maybe should we change the port range to
minPort <= port <= maxPort?

stevehu:
@jiachen1120 @NicholasAzar We are discuss the edge case now. In most of the cases, users won't use dynamic port if the minPort and maxPort is equal. They would assign the port directly without dynamic port enabled. That being said, we cannot predict user behaviors. In the case that minPort is equal to maxPort, we should try to bind to it and quit with error if that particular port is not availabe. If that means we have to break the previous behavior, then let's break it as I don't think anybody is depending on the behavior. I think we should change the port range to minPort <= port <= maxPort as @jiachen1120 suggested.

PR #419, opened 2019-03-03
just improvements

jefperito:
Hello guys, i am pushing these "improvements" of code


stevehu:
@jefperito I saw you have added audit.yml and handler.yml in the src/test/resources folder. Where are these files used? For the AuditConfig, is the default config file in src/main/resources/config/audit.yml loaded? 

I am still reading the PowerMock documentation and found that it still doesn't support JDK11 and JUnit 5 yet. I am a little bit concerned to use it as it is working on the bytecode level and might have compatibility issue when we upgrade other testing dependencies or JDK. Could you please point out where we have to use PowerMock as Mockito is not enough? Thanks.

jefperito:
Hello guys, 

sorry for delay, i've been very busy recently

@ddobrin Ok!

@stevehu i'm using PoweMock to test some cases where mockito does not fit, e.g. test static methods, and about the new branch: Ok!

stevehu:
@NicholasAzar @ddobrin @jiachen1120 What do you think to introduce PowerMock? @jefperito Could you please point out where the audit.yml and handler.yml in the src/test/resources are referred? Thanks.

ddobrin:
Powermock has better access to static methods, static initializers, private methods, things which could be useful.

If there is a compelling case to introduce it, I am not against it.

jefperito:
removed those yml files, i committed a mistake, just forgot delete the files because tests dont use anymore

stevehu:
@jefperito Thanks a lot for your help.

stevehu:
[just improvements](https://trello.com/c/4FfddXQn/117-just-improvements)

stevehu:
[just improvements (#419)](https://trello.com/c/KcMrQXxA/121-just-improvements-419)

PR #422, opened 2019-03-06
fixes #421 add last handler to Handler class for handler chain inject…

### LINKED ISSUE #421: add last handler to Handler class for handler chain inject in the beginning
While trying to integrate Light-4j with Spring Boot, we are injecting the handler chain in front of the Spring Boot ServletInitialHandler to address cross-cutting concerns at HTTP core level for Spring Boot Controllers. In this use case, we need to define the handler chain in handler.yml without the last handler. The last handler is passed in by the Spring framework when injecting OrchestratorHandler. Here is what we need to do. 

* add lastHandler to the Handler class
* add another constructor for OrchestrationHandler to accept a lastHandler parameter
* if next is null and lastHandler is not null, call the nextHandler.handleRequest

stevehu:
… in the beginning


NicholasAzar:
#421

Issue #420, opened 2019-03-04
add TTL to the consul check register for ttlCheck option


stevehu:
There are three options to allow Consul to update the health of individual service instance and they are defined in the consul.yml file. For most of the cases, we encourage users to use httpCheck as it is well tested and very reliable. However, for certain situations, the service instance cannot be accessed by the Consul cluster/agent, then ttlCheck might be used. Instead of letting Consul call your service, the service will call Consul instead in this case. 

For the other two options, httpCheck and tcpCheck, only one interval needs to be sent to the Consul so that Consul will make calls based on this interval. For ttlCheck as the call is initialized by service itself, we actually need two values to control how the health check works. 

TTL: This is the value that is sent to the Consul to indicate the timeout when Consul consider the service is not healthy. 

Interval: The service will access the check endpoint of Consul at this interval. 

The rule is that the TTL must be equal or greater than the Interval. This allows Consul to received a call before the TTL is reached. 

In the current implementation, we are getting the interval from consul.yml and use that as TTL value to send to Consul for the check register. We also use the same value to trigger the calls to consul from the service instance. As there two values are the same, we don't have any buffer in the case the check request is not handled in time. 

The recommendation is to add another ttl value for the consul.yml only for ttlCheck option and make this value bigger than the interval in the same config file. 

For example, if we set the interval to 10s, we can set the TTL 15s. So the Consul server will check if there is a call from the service every 15s and the service will call to the Consul very 10s. 

https://www.consul.io/api/agent/check.html

BalloonWen:
The actual reason is the "checkInterval" setting in consul.yml doesn't work for TTL, the heartbeat for TTL was not depended on that.
Now: the "checkInterval" will work for all TCP, HTTP, TTL in registration to Consul, but when TTL model, the heartbeat to send to Consul will be 2/3 of "checkInterval"

stevehu:
@BalloonWen Good findings. I think your solution is much better than putting two values into the consul.yml I originally thought. Thanks.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [add TTL to the consul check register for ttlCheck option](https://trello.com/c/n032XdUL/119-add-ttl-to-the-consul-check-register-for-ttlcheck-option)

kumasath:
Hi Steve,
As discussed, Consul registry failed with 2 pods.  Please find the stacktrace and consul.yml file below.
This issue can be simulated in Openshift container. Thanks

[...]

PR #428, opened 2019-03-18
Feat/light 4j/#420 fixed TTL check

BalloonWen:
fixed issue: https://github.com/networknt/light-4j/issues/420


stevehu:
[fixed issue for consul's ttl check, now the ttl interval will be still the same as "checkInterval", but the heartbeat will be 2/3 of "checkInterval" (#428)](https://trello.com/c/HZZoozfU/122-fixed-issue-for-consuls-ttl-check-now-the-ttl-interval-will-be-still-the-same-as-checkinterval-but-the-heartbeat-will-be-2-3-of)

Issue #427, opened 2019-03-08
Retrieve symmetric key for config decryption


jiachen1120:
@dz-1 has already implemented Integrate decryption module in all retrievals of config https://github.com/networknt/light-4j/issues/413

But the symmetric key is hard coded as framework name "light" in AESDecryptor. We need to make it configurable

processes need to be defined:
Deployed in VM:
Key retrieved by user input on server start
Deployed in CaaS:
Key retrieved by environment variable

stevehu:
The best way to make it work is through [light-config-server](https://doc.networknt.com/service/config/) as it defines a process to split the key between the service owner and the DevOps engineer. No single person knows the key in full but the config server and service instance knows the full key to enable the encryption and decryption.

jiachen1120:
@stevehu Hey. I think this requirement might apply to users who don't want to use the config server. In the absence of a config server, perhaps relying solely on config modules can lead to potential security issues? Since the password for decryption is fixed

jiachen1120:
@stevehu Also would like to ask you a question about the config server, Is the config got from config server is decrypted?

stevehu:
@jiachen1120 I agree that we need to have a solution for standalone applications; however, it would be very hard to just use the environment variable and we need a way to get the key inside the application itself. If you have some ideas already, please write an RFC so that we can discuss. 

When using the config server, the values should be encrypted by the config server and decrypted in the application. The key in the application is constructed by something set up in the application and an environment variable provided by the DevOps engineer.

PR #429, opened 2019-03-20
Feat/#427 config decryption

jiachen1120:
Hey! 
This PR is based on @dz-1 PR, https://github.com/networknt/light-4j/pull/414. Please merge that PR before this one :)

This PR includes the implementation of two new decryptor:
1. AutoASEDecryptor -- can retrieve configuration file decryption password from environment variables
2. ManualASEDecryptor -- can retrieve configuration file decryption password by stdin when server startup

This two implementation can be configured in the config.yml, more detailed can be referred from the RFC. Look forward to your review!

Related RFC: https://github.com/networknt/light-rfcs/blob/feat/%23427-config-decryption/light-4j/0018-retrieve%20symmetric%20key%20for%20config%20decryption.md

Related issue: https://github.com/networknt/light-4j/issues/427


dz-1:
The differences among the three classes, ASEDecryptor, AutoASEDecryptor, and ManualASEDecryptor, are minor. They are different only in the way of acquiring the password. 
The method `init()` doesn't have to be static, I think. We can move the common code to a common abstract class with an abstract `init()` method? 
Or, we can simply restructure `ASEDecryptor` a little bit and use it as the parent class of AutoASEDecryptor and ManualASEDecryptor.

dz-1:
Or, maybe use `Strategy pattern` to abstract away the difference.

miklish:
Also, shouldn't this be AES not ASE?

jiachen1120:
@dz-1 Hey, I restructured ASEDecryptor  and use it as the parent class of AutoASEDecryptor and ManualASEDecryptor. @miklish Yes, it should be AES.

NicholasAzar:
@jiachen1120 could you please take a look at the conflicts?

jiachen1120:
@NicholasAzar Sure, will work on it!

NicholasAzar:
[config decryption](https://trello.com/c/1Ilpssv9/128-config-decryption)

jiachen1120:
@ddobrin Hey, I changed the package name from utility to decrypt is because @dz-1 moved the decryptor to decrypt package in PR: https://github.com/networknt/light-4j/pull/414. And this feature would be release in 1.6x, same as: https://github.com/networknt/light-4j/pull/414#issuecomment-475755465.

Sure, I will document this feature in config part to specify how to use this feature.

ddobrin:
@jiachen1120 : please note that 1.6.x still indicates backwards compatibility, according to https://semver.org/.

As the master branch already contains changes which development teams wish to use in future projects, teams would have to make the changes imposed by this PR as well. 

I do understand that #414 is imposing this, however I would recommend that users are informed/notified or this issue addressed, as 1.6.x would still indicate backwards compatibility

dz-1:
@ddobrin We moved the interface from `util` to `decryptor` because the `util` module has too many dependencies. Since our changes are in the `config` module, there will be circular dependencies if we use the interface as is.
If this change is a concern, we would have to move the decryptor  related changes to later releases.

@stevehu Please advise.

stevehu:
We want to follow the rules from https://semver.org/ whenever possible. That is why we still keep the secret.yml in the master branch although it can be removed with the new feature supports decryption in every config file. When we talk about compatibility, we need to keep the scope in perspective, otherwise, we will reach version 100 very soon with all the new features introduced on a daily basis. 

The entire platform can be divided into core and new features. I think for the same minor version, we need to ensure that we won't break the core which most users are using. For new features introduced to the platform, it needs time to be stabilized. Since not a lot of users are using these features, we can correct the API mistakes within the same minor version. However, we need to clearly document the impact and guide users to upgrade in the release notes and documentation. 

For this particular decryptor, it was my mistake when it was introduced as I didn't think to put it into the config module but as a utility to call. It is much better to put it into the config module so that it is transparent for users. As I know, there are not a lot of users are using the encryption and decryption, so changing the package name and move the configuration from service.yml to config.yml shouldn't impact the majority of users but only some enterprise users who have restricted security policies. It happens that we have direct supporting channels with these users and we can help them to upgrade. 

This is just a starting and we might face similar situations very often down the road. I will try to facilitate a discussion internally and document the detail rules in our process.

stevehu:
@ddobrin I think your clients are impacted the most. Could you please review this PR and the document update by @jiachen1120 to gauge the impact and give us your feedback? Your evaluation will help us to perfect the document and guide us to write the release notes for other users.

ddobrin:
@jiachen1120 @stevehu : I will make the changes for my client and we'll move this forward.

I would like to recommend that we remove the old packaging from all code and config classes, to have a uniform 
For example, we should update: `/light-4j/data-source/src/test/resources/config/service.yml`

For the documentation, I would specify that the secret.yml approach is still available and that this was the approach:
secret.yml
   pwd : CRYPT eiuewyiweyiew

and the new approach is:
  ....
where config.yml looks like:
...
decryptorClass:
  com.networknt.decrypt.AESDecryptor

I would also mention that by default, the com.networknt.decrypt.AESDecryptor is wired into the config,yml available under light-4j/config/src/main/resources/config

Developers should be able to glance how it worked, how it works now and the available defaults

PR #446, opened 2019-04-01
fixes #445 update client module to access signing service

### LINKED ISSUE #445: update client module to access signing service
Jwt token can be used for security and it can be used for the signature. Need to add one more entry in the client.yml and add API to interact with signing service.

stevehu:


stevehu:
Document update can be found in https://github.com/networknt/light-doc/issues/60

Issue #437, opened 2019-03-29
Cache unparsed request body (configurable)


ddobrin:
The BodyHandler (light-4j/body) currently parses the body of the request and caches the parsed body as an attachment in REQUEST_BODY.

There are use cases when the "unparsed" body is useful for some operations and avoids a serialization effort.

The body.yml needs to be augmented with a property, with a default value set to "false".
Teams which will use it will set it to true as per their use case.
...# Enable body parse flag
enabled: true
...# cache request body
cacheRequestBody: false

The BodyHandler would have to be augmented and should save, if the above flag is set to true, the unparsed body as a String, as an attachment in REQUEST_BODY_STRING.

NOTE: light-codegen already parameterizes the body.yml at this time, therefore no changes are required in the codegen utility

jiachen1120:
Hey! @ddobri @stevehu I encounter a problem when trying to implement caching unparsed request body when the content type is "multipart/form-data" and look forward to your helps.  

Due to the extended problem of undertow, the only way to parse the form-data is using the method formParserFactory.createParser(exchange), which means we need to reuse the input stream inside the undertow exchange. However, the undertowInputStream didn't override the reset() method and cannot be reused. So, after caching the unparsed body from the undertowInputStream, we cannot parse the body from the exchange again since the input stream already closed. 

It seems the only way to solve it is to write an ourselves form-data parser. Do we need to do that? Or, just implement the caching for Json format? Since I think there is nearly no use cases for unparsed form-data body.

ddobrin:
@jiachen1120 : If you consider that this issue can not easily be addressed, then writing an additional parser is not warranted.
There are a couple of use-cases for it, at least the ones which I am aware of.

Please reassign the issue to me and I will close it in the next few days, after investigating a couple of different ideas. 
You can label it as "to be analyzed" or something you see fit

stevehu:
@ddobrin @jiachen1120 I guess the use cases you are talking about are for JSON only. In my opinion, there is no need to keep the original string for form-data as there is no original string at all. As our body parser differentiate body with the content type, we can just keep the original string for the `application/json`. What do you think?

jiachen1120:
@stevehu @ddobrin Thanks for your help!!!

I just created a PR which included: 
1. implementation of caching application/json request body string
2. refactored the bodyHanlder
3. test cases

I agree @stevehu opinion that there is no need to keep the original string for form-data as there is no original string.

PR #447, opened 2019-04-01
Cache unparsed request body (configurable/ for application/json request)

jiachen1120:
Related issue: https://github.com/networknt/light-4j/issues/437

Currently, the unparsed body can be cached into exchange attachment when the content type is "application/json"

Beside that, I refactored the bodyHandler a little bit to reduce the code redundancy and added test cases for the caching body string.

Look forward to your review!

### DIFF
--- a/body/src/main/java/com/networknt/body/BodyConfig.java
+++ b/body/src/main/java/com/networknt/body/BodyConfig.java
@@ -21,6 +21,7 @@ package com.networknt.body;
  */
 public class BodyConfig {
     boolean enabled;
+    boolean cacheRequestBody;
 
     public BodyConfig() {
     }
@@ -31,5 +32,13 @@ public class BodyConfig {
 
     public void setEnabled(boolean enabled) {
         this.enabled = enabled;
+    }
+
+    public boolean isCacheRequestBody() {
+        return cacheRequestBody;
+    }
+
+    public void setCacheRequestBody(boolean cacheRequestBody) {
+        this.cacheRequestBody = cacheRequestBody;
     }
 }
--- a/body/src/main/java/com/networknt/body/BodyHandler.java
+++ b/body/src/main/java/com/networknt/body/BodyHandler.java
@@ -2,7 +2,7 @@
  * Copyright (c) 2016 Network New Technologies Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
- * You may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


stevehu:
@jiachen1120 I have made two updates: 

* Remove the enabled check as the flag is used to control if the entire handler is enabled in the handler chain. 
* Change the type of the attachment to String from Object so that users won't need to cast it.

Please review and let me know if the changes are proper. Thanks.

jiachen1120:
@stevehu Good morning! Thanks for the updates!

In my opinion, maybe the entire handler should be disabled when both `enabled` and `cacheRequestBody` is disabled. 

Otherwise the user will not be able to get only the unparsed request body. There are only three cases here, either attach the parsed body, or attach both the parsed and unparsed request body, or nothing. But if check two flags, there are four cases:


  | body.yml | Expected Result 
-- | -- | -- 
1 | enabled: true  cacheRequestBody: false | only parsed body is attached to exchange, no unparsed body is attached 
2 | enabled: true cacheRequestBody: true | parsed body is attached to exchange. unparsed body is attached If the content type is "application/json". 
3 | enabled: false cacheRequestBody: true | no parsed body is attached, unparsed body is attached If the content type is "application/json".
4 | enabled: false cacheRequestBody: false | neither parsed and unparsed is not attached 

Or, Perhaps because there is no such use cases to get the unparsed body only？

stevehu:
@jiachen1120 The enabled flag is part of the middleware handler interface. If the flag is false, it simply means that the handler is skipped during the runtime. This gives users an opportunity to disable a middleware handler without removing it from the handler chain definition. Given this usage of the flag, the case 3 and 4 are not valid as the handler is disabled already.

jiachen1120:
@stevehu Yes, so do we need to provide another flag to control the body parsing? before, we do not need it because the handler only has one function: parse the body and attach it to exchange. disable the handler means disable this function.

But now we have two function in this handler. Should we control them separately? 

For example:
enabled: // control the entire handler
parseBody: // control whether parsing the body
cacheRequestBody: // control whether attach the original request to attachment

stevehu:
I think the current implementation should be fine unless we have new requirement from our users.

jiachen1120:
@stevehu I got it. Thanks 👍

NicholasAzar:
[Cache unparsed request body](https://trello.com/c/pKnMvlKp/130-cache-unparsed-request-body)

PR #453, opened 2019-04-05
Add logging module for light 4j   RFC#29

sreenicibc:
New logging module created for light 4j to see and change the logging levels .


NicholasAzar:
@sreenicibc could you resolve the conflicts in this?

sreenicibc:
I have resolved it, Please check now.

NicholasAzar:
Looks fine to me.. pending rfc approval/merge

stevehu:
I think we need to have a security mechanism in place, otherwise, it might be a security hole opened for an attack. What do you thing?

NicholasAzar:
@stevehu i'm not sure that's necessary at the moment, as what's added are ultimately just a few handlers that can be secured (or not) just like any others at the developers discretion..
What i was thinking was a more basic security method for this simple task that can be added later.

stevehu:
@NicholasAzar Let's separate all the handlers to business handlers and technical handlers. Business handlers are built by the users and we are expecting the user to enable oauth2 to protect them in most of the cases. For technical handlers, they are part of the platform features like server info, health check, and logging level. These endpoints must be enabled in the handler.yml with security handler in the request/response chain. Ideally, it should be the same as the business handlers so that users won't need to maintain two different authorization subsystem. Before we wire it into the handler.yml in the light-codegen, there is no risk for the users. However, I think we need to come up with an idea in design to allow security to be enabled. It might be just a document to guide users on how to use the feature securely.

NicholasAzar:
@stevehu could you please let us know what you're looking for? Is it just documentation? Do you have any samples for securing the other "technical handlers"?

stevehu:
I am looking for a strategy and document at the moment. And the real impl when we enable the feature.

NicholasAzar:
@sreenicibc would you be able to add some documentation to light-doc in the "concerns" section outlining how to use, secure, and configure these handlers?

sreenicibc:
I have added the logging.md file in concerns section and created the PR .

stevehu:
[Add logging module for light 4j RFC#29](https://trello.com/c/lWmErCxi/248-add-logging-module-for-light-4j-rfc29)

Issue #412, opened 2019-02-28
Support configure for allowUnescapedCharactersInUrl in undertow.


hebutliu2010:
The old version of ligth4j(1.3.5) support non-ascii characters for query parameter, but new version of ligth4j dependent on new version undertow, this feature is no longer supported.
https://github.com/undertow-io/undertow/commit/d63b767010133d574196740892a845c1375e8bed#diff-4d93e9a9cf8b37e2bc2b882ae2bca189

We can add configure for allowUnescapedCharactersInUrl.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Support configure for allowUnescapedCharactersInUrl in undertow](https://trello.com/c/l6VJx5cr/168-support-configure-for-allowunescapedcharactersinurl-in-undertow)

hebutliu2010:
Thank you very much.

stevehu:
@hebutliu2010 I have added the following comment on the document site and you need to be aware that there is a security risk to enable this feature. 

```
 # Flag to set UndertowOptions.ALLOW_UNESCAPED_CHARACTERS_IN_URL.
 # Please note that this option widens the attack surface and attacker can potentially access your filesystem. 
 # This should only be used on an internal server and never be used on a server accessed from the Internet. 
 allowUnescapedCharactersInUrl: true

```

PR #462, opened 2019-04-11
Added new server option ALLOW_UNESCAPED_CHARACTERS_IN_URL

jiachen1120:
related issue: https://github.com/networknt/light-4j/issues/412

Added new server option `allowUnescapedCharactersInUrl` to support new feature of undertow. It can be configured just like other server options (ioThread, workerThread...). The default value of it is `true`. The documentation will be updated later.

### DIFF
--- a/server/src/main/java/com/networknt/server/Server.java
+++ b/server/src/main/java/com/networknt/server/Server.java
@@ -244,6 +244,7 @@ public class Server {
                     // HTTP/1.1 requests, as it is not required
                     .setServerOption(UndertowOptions.ALWAYS_SET_DATE, config.isAlwaysSetDate())
                     .setServerOption(UndertowOptions.RECORD_REQUEST_START_TIME, false)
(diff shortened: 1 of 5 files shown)


NicholasAzar:
I think we should default this to false right? At least that's what undertow does..

jiachen1120:
@NicholasAzar Hey! The old version of light-4j supports non-ascii character in Url. This feature disappears due to the update of undertow. Perhaps for backward compatibility, we need to keep the same performance as the previous version, so the default value of this flag maybe needs to be true. If we use the default value of undertow, it may cause some incompatibility, although I don't think anyone will use some rare symbols in the URL. for example: Ã¥... What do you think?

NicholasAzar:
For the 1.6.* branch, i would keep it as false to match undertow. If we want to change this to match how it was working previously, maybe we should just do it for people using the 1.5.* branch?

jiachen1120:
@NicholasAzar I agree! Will change it 👍

PR #475, opened 2019-04-17
fixes #474 register light-hybrid-4j services in Server module

### LINKED ISSUE #474: register light-hybrid-4j services in Server module
For Rest and GraphQL service, there is one instance of server per service so it needs to register the server instance to the Consul only. However, one light-hybrid-4j server can have numeric instances of services. We need to register all the services to the Consul so that a client can loop up a particular service by Id. 

In the light-hybrid-4j framework, a startup hook is responsible for merging all services together and it has all the service Ids available. We cannot call the Server method to register at the moment as the startup hooks are called before the bind is invoked. The idea is to populate an array list of service Ids in the Server instance and later on register this list of ids after the server instance is registered. 

For the Server, we need to extract the register into a method and also switch heartbeat after all services are registered.

stevehu:


santoshaherkar:
@stevehu I see that the static server config variable is added back again in Server class. We had removed that variable to support config server solution. Config Loader is triggered later in the class which fetches config values and puts it in Config.configCache. So, all the configs should be read after that step & it wont work as expected if we read configs before this step. 
Other reason was, config is cached in Config class & it should be maintained there only. If we keep static variables for configs in other classes, its as good as cashing the config in those classes. In that case, if the configs are refreshed in Config.configCache, other classes (using static configs) will continue using old values, which is not good. 
If its absolutely required to keep that static config variable in Server class, we have to delay its  initialization at least; otherwise config server solution will not work quite well.

stevehu:
@santoshaherkar I didn't realize that there is a timing issue when l loading the static server configuration. I rollback the changes because I was reported by multiple users that their applications are broken after upgrade to 1.6.0 and they have to downgrade back to 1.5.x. We have too many places in our application and light-codegen to rely on the Server.config. Unless we fix all references in the platform and ask users to regenerate their code, we cannot remove the config. We need to have an internal discussion for the solution.

santoshaherkar:
@stevehu As discussed on the phone call, I have made the changes to make it backward compatible and to keep config server solution working as well.

stevehu:
@santoshaherkar I have reviewed the code and it looks good. Thanks a lot for your help.

santoshaherkar:
@stevehu when are you planning to merge this PR?

stevehu:
@santoshaherkar Thanks a lot for your help. I am planning to merge it before the 1.6.1 release. I would like another member to review it so that we can ensure that no issues introduced by this PR.

Issue #480, opened 2019-04-22
client module enhancement to use CompletableFuture to handle two or more threads  asynchronous computation


chenyan71:
client module implement to use CountDownLatch to handle threads wait. 
Add some new methods to java CompletableFuture to handle two or more threads  asynchronous computation

jefperito:
Is There any way create unit tests?

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [add new feature method for client module.](https://trello.com/c/bWKfqfOg/176-add-new-feature-method-for-client-module)

PR #481, opened 2019-04-22
fixes issue #480

chenyan71:
add new feature method for client module.

The method call wrap will be implemented on light-consumer-4j

### DIFF
--- a/client/src/main/java/com/networknt/client/http/Http2ClientCompletableFutureWithRequest.java
+++ b/client/src/main/java/com/networknt/client/http/Http2ClientCompletableFutureWithRequest.java
@@ -1,0 +1,58 @@
+package com.networknt.client.http;
+
+import io.undertow.client.ClientCallback;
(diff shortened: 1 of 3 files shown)


client/src/main/java/com/networknt/client/http/Http2ClientCompletableFutureWithRequest.java
```diff
@@ -0,0 +1,58 @@
...
+    Logger logger = LoggerFactory.getLogger(Http2ClientCompletableFutureWithRequest.class);
+
+    String requestBody;
+    public Http2ClientCompletableFutureWithRequest(String requestBody) {
+        this.requestBody = requestBody;
+    }
+
+    @Override
+    public void completed(ClientExchange result) {
+        new StringWriteChannelListener(requestBody).setup(result.getRequestChannel());
```
yibogithub: I may have forgotten the good reason. Why can't we factor out the common code?

PR #485, opened 2019-04-24
#240 timeout feature #241 circuit breaker

### LINKED ISSUE #240: Provide a handle to set connection timeout on ClientConnection object
The light4j library’s Http2Client uses non-blocking socketChannel but it doesn’t seem to have a connection time-out configured.
 
```
connection = client.connect(new URI(apiHost), Http2Client.WORKER, Http2Client.SSL, Http2Client.POOL, APIProtocol.equalsIgnoreCase(PROTOCOL_H2) ? OptionMap.create(UndertowOptions.ENABLE_HTTP2, true): OptionMap.EMPTY).get();
```
 
The connection.sendRequest can be wrapped in a Latch to set a read timeout.
 
connection.sendRequest(request, client.createClientCallback(reference, latch, requestBody));
boolean responseReceived = latch.await(timeout, TimeUnit.SECONDS);
 
The non-blocking setting allows an asynchronous way to connect and read/write separately, however, we should have a finite interval to stop trying to establish a connection to a rest service (getting a connection object).
 
The internal class socketChannel allows to set socket timeout; however, there is no direct handle to it. Please share your views and any available information.
 
FY Reference:
http://www.java-gaming.org/index.php?topic=17250.0
https://community.oracle.com/thread/1145435

### LINKED ISSUE #241: Add circuit breaker to the client module
Previously, we recommend using Hystrix for high volume service invocation. However, most of the times, it is overkill. It makes sense that we implement the pattern in the client module.

jefperito:
Configuration needed for circuit breaker (into client.yml):
```yml
request:
  errorThreshold: 2
  timeout: 3000 (in ms)
  resetTimeout: 7000 (in ms)
```
if parameters are not given, there are default values for each request configuration into ClientConfig.java:
```java
    public static final int DEFAULT_ERROR_THRESHOLD = 5;
    public static final int DEFAULT_TIMEOUT = 3000;
    public static final int DEFAULT_RESET_TIMEOUT = 600000;
```

Http2Client::getRequestService returns an async service with circuit breaker feature.

Http2ClientTest there are integration tests for each state of circuit breaker (open, half open and close)

In CircuitBreaker (resolves #241), uses CompletableFuture::open(time, unittime) API resolves #240 issue.


jefperito:
@ddobrin ok! I'll improve that way

stevehu:
@jefperito I have merged from the latest code from master and resolve the conflict. Please check out the latest, review and make the additional modification. Thanks.

jefperito:
@ddobrin it's done

PR #486, opened 2019-04-24
Fix/#482 consul service discovery caching

BalloonWen:


BalloonWen:
in LightCluster -> discovery(String protocol, String serviceId, String tag)
initially the urls will contain all the available service urls including all the tags
when tag is null, return all the discovered urls
when tag is "", return urls with empty tag
when tag is a non-empty String, will filter the urls by tag

Issue #491, opened 2019-04-25
direct registry stops working as it doesn't support env tag


stevehu:
We have an issue fixed in the service discovery that the tag is not part of the cache. https://github.com/networknt/light-4j/issues/482

After the fix, it seems that the direct registry stops working as the tag is not supported in it. I have fixed the test case for the light-router in the following issue to disable the tag in the header. 

https://github.com/networknt/light-router/issues/49

Also, I have received a report that the user application fails with a direct registry in the dev environment. 

We need to figure out why the direct registry fails and see if we can enhance it to mimic the feature of Consul Registry.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [direct registry stops working as it doesn't support env tag](https://trello.com/c/DlnQbUMX/183-direct-registry-stops-working-as-it-doesnt-support-env-tag)

PR #492, opened 2019-04-25
Fix/#491 direct registry with env

jiachen1120:
issue: https://github.com/networknt/light-4j/issues/491


stevehu:
That makes sense. Let's merge it and test it.

PR #501, opened 2019-05-02
Feat/default decryptor

jiachen1120:
Previous behaviour: when no decrypt class specify in config.yaml. No decryption apply

Change to: when no decryptor class specify in config.yml or no config.yml at all. Using AESDecryptor as the default. In this way. the backwards compatibility is maintained.


stevehu:
[Feat/default decryptor](https://trello.com/c/ertWmNy0/194-feat-default-decryptor)

Issue #502, opened 2019-05-06
Adding auto test for consul registry and discovery 


jiachen1120:
Currently, a mock consul may cannot fully cover the real usage cases of consul.  To integrate with something like this https://www.baeldung.com/docker-test-containers to start a real consul server in our tests so we can automate a bit more.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Adding auto test for consul registry and discovery](https://trello.com/c/KmqMQvwL/254-adding-auto-test-for-consul-registry-and-discovery)

PR #503, opened 2019-05-06
Feat/#502 consul integrate test

jiachen1120:
Enhance the registry test by integrate with https://www.baeldung.com/docker-test-containers, so a real consul can be established automatically when testing.

Since this integrate test depends on docker so it should be disabled all the time unless it is used.

Related issue: https://github.com/networknt/light-4j/issues/502


stevehu:
@jiachen1120 This is a very good idea to test with a real container; however, we cannot expect all developers to use docker on their development environment. We have some repositories that depending on the external containers for testing and normally we create an extra folder called integration along with test for integration test. All unit tests will be in the test folder and they don't rely on the container and all integration test cases are in the integration folder and they rely on the container. The following is an example in light-session-4j and the integration test won't be invoked when `mvn clean package` is called. 

https://github.com/networknt/light-session-4j/tree/master/redis-manager/src

ddobrin:
@jiachen1120 : while I agree that it is a neat idea, please be aware that developers at my client are building/testing in several APIs on Windows machines, on prem, in order to have access to certain resources.

In those cases, they don't use Docker, as they don't have it on their machines.

jiachen1120:
@stevehu Cool, I will migrate this test to an integrate folder

jiachen1120:
@ddobrin Hey, after migrating this test to integration test folder. They would not be invoked during mvn build. They can be used just by developer to testing the framework.

stevehu:
I looked at the test container project and it is very promising. It is much more convenient than our docker-compose. Good job!

server/src/integration/java/com/networknt/server/ConsulRegistryIT.java
```diff
@@ -0,0 +1,187 @@
...
+
+import static com.networknt.consul.ConsulConstants.CONFIG_NAME;
+import static com.networknt.server.Server.STATUS_HOST_IP;
+import static org.junit.Assert.fail;
+
+/**
+ * This test is an automatic test for real consul registry and discovery by using testContainers.
+ * Since it depends on docker so it should be disabled all the time unless it is used.
+ */
+@Ignore
```
NicholasAzar: Is there a better way to enable/disable then (what I presume) is manually removing the @ Ignore flag? What is the process for running full integration tests?

stevehu: We can enable the test case. If you don't want to execute the integration test cases, use `mvn clean package`. If you want to run all integration tests along with unit tests, then you can use `mvn clean install`.

NicholasAzar: Maybe we don't want to go that far in case someone wants to install the dependencies but doesn't have docker installed.. could we add a maven profile for running integration tests?

stevehu: Using maven profile is another option. The key point is we need to ensure that broader users can compile without the docker installed and our developers can trigger integration tests with a switch. I really have no idea which commands most developers using `mvn package` or `mvn install`.

jiachen1120: @stevehu Hey. I think `mvn install` would be more popular to use. So I updated the pom to disable the integration test execution when using `mvn install`.  

In this way,`mvn install` only run the unit test.
And the following commands are using for executing integration tests:
mvn verify -DskipITs=false
mvn integration-tests -DskipITs=false
mvn clean install -DskipITs=false

What do you think?

stevehu: I think it is a very good idea. mvn verify -DskipITs=false can execute integration tests when they are annotated as ignored?

jiachen1120: I just tested. If `@ignore` annotated inside testing code. There seems no way to execute them. I think we can remove all the ignore annotation in the integration test if the default value for skipITs is true.

stevehu: Got it. I like the idea. We need to make sure that everything is transparent to new developers.

NicholasAzar:
@jiachen1120 i'm getting failures when running the build locally.. is there something i'm missing?
```
[ERROR] Failures: 
[ERROR]   ConsulRegistryIT.testDiscoverCaching:194 expected:<...//192.168.85.121:495[90]> but was:<...//192.168.85.121:495[89]>
[ERROR]   ConsulRegistryIT.testDiscoverWithNonEmptyStringEnvTag:169 expected:<...//192.168.85.121:495[90]> but was:<...//192.168.85.121:495[89]>
```

NicholasAzar:
> @jiachen1120 i'm getting failures when running the build locally.. is there something i'm missing?
> 
> ```
> [ERROR] Failures: 
> [ERROR]   ConsulRegistryIT.testDiscoverCaching:194 expected:<...//192.168.85.121:495[90]> but was:<...//192.168.85.121:495[89]>
> [ERROR]   ConsulRegistryIT.testDiscoverWithNonEmptyStringEnvTag:169 expected:<...//192.168.85.121:495[90]> but was:<...//192.168.85.121:495[89]>
> ```

This passes for me now with command `mvn verify -P integration-test`

Issue #506, opened 2019-05-08
Make sure request/responses for consul, oauth, and api calls are logged with debug


jiachen1120:
In the case of consul I believe the request could be pieced together with the results of previous log statements, but we don’t currently output the entire request/response as a single entry.
For the OAuth calls and the ones to downstream APIs, although there are a few, on looking through the code our log statements are definitely more sparse… If required, the calls to the downstream APIs and their responses can be logged from your application code.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Make sure request/responses for consul, oauth, and api calls are logged with debug](https://trello.com/c/QKsRUkQl/250-make-sure-request-responses-for-consul-oauth-and-api-calls-are-logged-with-debug)

PR #507, opened 2019-05-08
Feat/#506 add debug logging

jiachen1120:
Related issue: https://github.com/networknt/light-4j/issues/506


client/src/main/java/io/undertow/client/http/Http2ClientCompletableFutureNoRequest.java
```diff
@@ -24,6 +24,9 @@ public void completed(final ClientExchange result) {
 
                     @Override
                     protected void stringDone(String string) {
+//                        if (logger.isDebugEnabled()) {
+//                            logger.debug("Service call response = {}", string);
+//                        }
```
NicholasAzar: Is there a reason we need to change the Undertow copied classes? Normally they would be under a different log level anyway..

jiachen1120: Once adding this logger...The travis test won't pass. Need further research to find the reason.

BalloonWen:
after did the cleanup in branch fix/#526-package-clean-up it seems working fine now.

Issue #512, opened 2019-05-14
Parameterized value in Config can not be overwritten 


ddobrin:
The Config module supports parameterized values.

In case a value is defined, and a default is provided, in any config file, the value can be overridden in values,yml

The following doesn't work:
server.yml
environment: ${server.environment:dev}

can be overridden by:
server.environment:test1

however can't be reset to null, as in:
server.environment:
or
server.environment: null

The behaviour is uneven, you can define the variable as:
environment: ${server.environment}
and set it to null as:
server.environment:

In this case, you can set it to null.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Parameterized value in Config can not be overwritten](https://trello.com/c/uz8ju0g9/213-parameterized-value-in-config-can-not-be-overwritten)

PR #516, opened 2019-05-15
Fix/#512 config overwritten

jiachen1120:
After the fix. users can inject `null` and `""` by using values.yml or environment variable

For example, in server.yml
```
environment: ${server.env:dev}
```
There are two ways to inject null or empty string:
1. Using values.yml
```
server.env:
// Setting empty string
server.env: ""
```
2. Using environment variable:
echo server.env=

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ dist/
 .project
 .classpath
 .settings
+.attach_*
 .metadata/
 *.iml
 *.log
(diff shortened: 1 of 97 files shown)


NicholasAzar:
Linking issue: #512

ddobrin:
@jiachen1120 : please update the documentation indicating that both "" and null are valid values.
thank you

Issue #521, opened 2019-05-27
Make the protocol of the consul client configurable


jiachen1120:
Currently, when using https for consul client, the protocol is default to http/2. Maybe we should make this configurable by providing a config entry in consul.yml
```
enbleHttp2: false
```

stevehu:
@jiachen1120 As consul supports HTTP/2 under HTTP/s natively, is there any reason we want a user to use HTTP/1.1 instead?

jiachen1120:
@stevehu when we upgraded the consul version it was no longer supporting our HTTP2 client using undertow. We tried to manually disable it and it worked (with https). The only way to disable HTTP2 without manual intervention right now is to use HTTP, which we can't do for a deployment to a server without a consul agent.. so we need it to be configurable

stevehu:
In this case, it makes perfect sense to make the HTTP2 configurable. Do you know why the new version doesn't support the HTTP/2 anymore? Have you opened an issue in consul? Which version of the Consul has this problem? I want to have a test on my server by upgrading to the consul to that version. Thanks.

jiachen1120:
@stevehu Good morning! The current consul version 1.2.2. We are trying to fix the issue. The problem is when using light-router to discover multiple service_id from consul multiple times. The process would stuck after the first success discovery.

stevehu:
I am using 1.2.2 in my testing cluster and I am planning to upgrade it to the latest soon. One thing I realize is that the HTTP/2 was only supported by Consul when HTTPS is enabled during the testing. If you are using a docker container for local testing, it won't work.

jiachen1120:
Hey. @BalloonWen and I test locally (local router and local services) by using a 1.4.4 docker container with HTTP/1. The behaviour is same as the HTTP/2.

jiachen1120:
Hey. @stevehu 

The consul issue can be found here:
https://github.com/networknt/light-4j/issues/523

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Make the protocol of the consul client configurable](https://trello.com/c/lyqltes8/246-make-the-protocol-of-the-consul-client-configurable)

stevehu:
In the previous implementation, HTTP/2 is true by default. With the enableHttp2 flag added to the consul.yml, the default becomes false if it is missing from the consul.yml

PR #522, opened 2019-05-28
Feat/#521 enbale http2 consul config

jiachen1120:
Related issue: https://github.com/networknt/light-4j/issues/521

Default value of enableHttp2 in consul.yml is false.

If no such entry in consul.yml, the value also be false.


stevehu:
@jiachen1120 Could you please update the light-doc for this new config field? 

https://doc.networknt.com/concern/consul/

PR #524, opened 2019-05-28
bug when using consul with http1.1

BalloonWen:
-after first lookup, do the block query, it will return the diff status.
-when first lookup, the consulindex is 0, it will do the quick lookup


stevehu:
@BalloonWen It should work for the consul client for this particular case. However, I think we should put the fix into the Http2Client so that we can handle HTTP/1.1 gracefully for all services who are leveraging the Http2Client.

Issue #519, opened 2019-05-23
Server Keystore should not fall back to default


jiachen1120:
Example:
User configures their HTTPS server keystore but puts in the wrong filename, when the server doesn't find it, it will default to the open source test keystore and will run normally.
This is a security issue.

A possible solution: migrate .keystore and .truststore from main folder to test folder

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Server Keystore should not fall back to default](https://trello.com/c/bE9KlEV8/256-server-keystore-should-not-fall-back-to-default)

PR #525, opened 2019-05-28
Fix/#519 keystore fall back

jiachen1120:
Related issue: https://github.com/networknt/light-4j/issues/519


server/src/main/java/com/networknt/server/Server.java
```diff
@@ -396,7 +396,7 @@ protected static KeyStore loadTrustStore() {
                 throw new RuntimeException("Unable to initialise TrustManager[]", e);
             }
         } else {
-            logger.warn("Unable to find server truststore while Mutual TLS is enabled. Falling back to trust all certs.");
+            // Mutual Tls is disabled, trust all the certs
             trustManagers = TRUST_ALL_CERTS;
```
NicholasAzar: When Mutual TLS is enabled, I think we should still display that warn message.

jiachen1120: Hi!

This function is invoked like:
```
if (getServerConfig().isEnableTwoWayTls()) {
     trustManagers = buildTrustManagers(loadTrustStore());
} else {
      trustManagers = buildTrustManagers(null);
}
```
Therefore, if the Mutual TLS is enabled, the `logger` line would not be executed. 

For `loadTrustStore()` method, it would not return `null` due to the changes in Config.class. When server.truststore cannot be found, an ConfigException would throw, which contains the reason: "Unable to load 'server.truststore', please provide the keystore/truststore matching the configuration in client.yml/server.yml to enable TLS connection."

Beside the exception, the error also be logged as : 
```
logger.error("Unable to load truststore " + name, e);
```

The keystore has the same behaviour.

Do you means that when enable TMLS without truststore, just warning and trust all the cert instead of throw an exception?  Would that lead to some security problems?

NicholasAzar: > Do you means that when enable TMLS without truststore, just warning and trust all the cert instead of throw an exception? Would that lead to some security problems?

No, that is how it works now but I don't think it's a good practice.

Currently there still could be issues if users choose different extensions for their keystrokes, like .jks for example. Could we make this independent of what the filename for the keystone actually is?

jiachen1120: Thanks for the reply!
This has been fixed. I put the keystore and truststore loading into TlsUtil.class for centralized management. In this way, the client and server will have consistent performance when loading the keystore and truststore. When loading a keystore or truststore, regardless of any extensions, an exception is thrown as long as the input stream cannot be obtained by its name.

config/src/main/java/com/networknt/config/Config.java
```diff
@@ -424,7 +422,7 @@ private InputStream getConfigStream(String configFilename, String path) {
                 logger.info("Unable to load config " + Encode.forJava(configFilename) + ". Looking for the same file name with extension yaml...");
             } else if (configFilename.endsWith(CONFIG_EXT_YAML)) {
                 logger.info("Unable to load config " + Encode.forJava(configFilename) + ". Looking for the same file name with extension json...");
-            } else {
+            } else if (configFilename.endsWith(CONFIG_EXT_JSON)) {
                 System.out.println("Unable to load config '" + Encode.forJava(configFilename.substring(0, configFilename.indexOf("."))) + "' with extension yml, yaml and json from external config, application config and module config. Please ignore this message if you are sure that your application is not using this config file.");
             }
```
NicholasAzar: Should the error message change along with this condition change?

BalloonWen: I think it's good, `else if (configFilename.endsWith(CONFIG_EXT_JSON))` equals to `else` in this case, because currently only three types of extensions.
the error message means we cannot find any config with these extensions.
The place to log this is weird, but it does its functionality.

PR #549, opened 2019-06-17
Add OpenTracing support for observability

stevehu:


stevehu:
I am still writing the distributed tracing tutorial and testing the Http2Client passing the context between services. The plan is to extend the service discovery tutorial to add one more step.

stevehu:
If you look at the JaegerHandler, it does exactly what you have described. If a context is passed in, then use that as the parent context, otherwise, create a new context.

PR #553, opened 2019-06-19
fixes #552 get client_secret from client.yml instead of secret.yml

### LINKED ISSUE #552: get client_secret from client.yml instead of secret.yml
Previously, we only have secret.yml decrypted during loading for encrypted values. Now, the logic is built into the config module so decryption is done automatically. It doesn't make sense to separate secret values into another config file. This issue is only to address the OAuth 2.0 interactions in the client module first. For backward compatible, we try to load the secret from client.yml and it is not there, then fallback to secret.yml.

stevehu:


client/src/test/resources/config/client.yml
```diff
@@ -14,34 +17,72 @@ tls:
...
+      #capacity of caching TOKENs
+      capacity: 200
+    # The scope token will be renewed automatically 1 minutes before expiry
+    tokenRenewBeforeExpired: 60000
+    # if scope token is expired, we need short delay so that we can retry faster.
+    expiredRefreshRetryDelay: 2000
+    # if scope token is not expired but in renew windown, we need slow retry delay.
+    earlyRefreshRetryDelay: 4000
+    # token server url. The default port number for token service is 6882.
+    server_url: https://localhost:6882
```
BalloonWen: changing port may cause test fail

stevehu: I just blindly copied and it was my mistake. Thanks for pointing it out.

PR #555, opened 2019-06-19
fixes #554 move the consul token to the consul.yml from the secret.yml

### LINKED ISSUE #554: move the consul token to the consul.yml from the secret.yml
For backward compatibility, fallback to secret.yml if it doesn't exist in the consul.yml

stevehu:


stevehu:
https://github.com/networknt/light-4j/pull/553

Issue #547, opened 2019-06-13
Deprecate certain public APIs across the framework


BalloonWen:
We've had a few instances where users of the framework would access public variables or methods that should not have been made accessible.
We should look through all public APIs to ensure that they are deprecated if they should not be accessed externally and ensure we have a decommissioning date included.

stevehu:
Totally agreed. Let's mark them deprecated in 1.6.x branch and remove them in the 2.0.x branch.

jefperito:
i can help you if you wish for that.

stevehu:
@jefperito I have invited you to the networknt slack channel and you can chat with @BalloonWen there. There are a lot of modules and repositories that need to be reviewed. Thanks.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Deprecate certain public APIs across the framework](https://trello.com/c/kWXCQYYN/319-deprecate-certain-public-apis-across-the-framework)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Deprecate certain public APIs across the framework](https://trello.com/c/Cl1UCazn/340-deprecate-certain-public-apis-across-the-framework)

PR #561, opened 2019-06-20
Fix/#547 deprecate apis - client module

BalloonWen:
related to issue: https://github.com/networknt/light-4j/issues/547

To avoid too many changes in one PR, @stevehu suggested that we should divide changes into smaller pieces, each PR will contain one module or multiple modules depend on how many classes are affected. It's more convenient for us to review, to trace or to revert.

Also, I need some suggestions:
- for ClientConfig.java, when we call "getTokenConfig()", "getSecretConfig()", should we return a map or we created a config model for each section?
- in Http2Client.java, there are too many overloads of connect(), and I believe each method has its specific use case for some users, we may need to provide a better way. For now, I don't want to touch any of those.


NicholasAzar:
@BalloonWen please ensure this is merged to the jdk 8 branch.. we can't support jdk 11 yet

PR #562, opened 2019-06-20
Merge to 1.6.x/#547 deprecate apis

BalloonWen:
related to issue: https://github.com/networknt/light-4j/issues/547

To avoid too many changes in one PR, @stevehu suggested that we should divide changes into smaller pieces, each PR will contain one module or multiple modules depend on how many classes are affected. It's more convenient for us to review, to trace or to revert.

Also, I need some suggestions:
- for ClientConfig.java, when we call "getTokenConfig()", "getSecretConfig()", should we return a map or we created a config model for each section?
- in Http2Client.java, there are too many overloads of connect(), and I believe each method has its specific use case for some users, we may need to provide a better way. For now, I don't want to touch any of those.


stevehu:
@BalloonWen I am reviewing the PR and found the changes are massive. As the codebase between 1.6.x and 2.0.x are different. I am wondering if you could submit another PR for the 2.0.x branch? Thanks.

BalloonWen:
> @BalloonWen I am reviewing the PR and found the changes are massive. As the codebase between 1.6.x and 2.0.x are different. I am wondering if you could submit another PR for the 2.0.x branch? Thanks.
@stevehu 
I think I already created PR for 2.0.x https://github.com/networknt/light-4j/pull/561
looks right now there are some conflicts, will resolve that later.

stevehu:
To support some of the users to go to production with the 1.6.x branch, we are going to wait until 1.6.5 release and branch out to 1.7.0 to merge this PR. The corresponding PR for 2.0.x has been merged already. This only for the JDK8

Issue #514, opened 2019-05-15
make the sanitizer handler context aware by config


stevehu:
The current sanitizer handler blindly applies the JavaScriptSource encoding for the body and it might not be the right thing to do based on the context of the API. The library was designed for JSP and server-side rendering with a clear context about the data it is trying to encode. In the API world, we don't have the context available. In order to apply the right encoding for the right API, we need to allow users to pass in a context from the sanitizer.yml so that the right encoding will be applied for the body of the request. 

We also need to explore if there are new libraries available to replace the https://github.com/OWASP/owasp-java-encoder

https://github.com/networknt/light-4j/issues/513

jefperito:
Do you have any idea how to define a context?

i did not find stable and active libraries for XSS in java.

stevehu:
I think it is the service developer's responsibility for defining the context in the configuration. Every service is dealing with some type of data and the developer knows which type of encoding is needed.

ddobrin:
I agree that the development team should be allowed to set the encoding. 

My client has a service they wish to build where they were asking for this

stevehu:
When doing sanitization for the XSS, we need to know what kind of data we are dealing with and apply certain type of encoding based on the data type and attribute as well as the position in the json. As a framework developer, we need to consider all possible scenarios but we cannot predict the use cases. We need the application developers to provide additional info to apply XSS encoding and we call it context since the app developers know what kind of data they are dealing with. The context can be passed to the sanitizer hander through the sanitizer.yml config file so that we know what kind of encoding is best for the application.

jefperito:
we can create a new attribute in config yml, where receives a list of json properties, so apply encode only these properties, or a little bit elaborated a list of [jsonpath](https://restfulapi.net/json-jsonpath/).

stevehu:
Good idea. That gives us even fine-grained control on which field to be encoded and what kind of encoding is needed.

jefperito:
i've started a new branch for resolve this one

https://github.com/networknt/light-4j/tree/514-sanitizer-new-config-properties

jefperito:
how about developers configure your own encoding? 

I'm thinking of allowing developers to configure their encoding engines implementing this interface i've created:

https://github.com/networknt/light-4j/blob/514-sanitizer-new-config-properties/sanitizer/src/main/java/com/networknt/sanitizer/enconding/Encoding.java

user to register that class (similar to ModuleRegistry), so at startup of sanitizer module, it recovery and instance that one.

[...]

PR #563, opened 2019-06-20
514 sanitizer new config properties

jefperito:
#514 
attributesToAppreciate: attributes for apply encoding (ignore everything else)
attributesToIgnore: attributes for ignore encoding (apply everything else)


stevehu:
@jefperito I have manually synced the master to the feature branch as the base has been changed from jdk8 to jdk11. One question: why do you name the positive list as attributesToAppreciate instead of attributesToEncode?

jefperito:
Sure, it's seems a better name to this purpose

jefperito:
done, that commit adds an integration test for encoding extended by client.

jefperito:
i don't know why that build step is broken:

"No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

stevehu:
It seems we have some problem with Travis. The local build is OK always and I guess the multiple threads test in client module is not working on Travis as it only allocates one thread per build. There might be other reasons but I don't know. It is nothing to do with this PR.

Issue #523, opened 2019-05-28
Consul connection freezing bug


jiachen1120:
In the process of consul lookup service. There is only one connection per consul client. This means that all lookup threads need to share this connection. And once a service is found successful. This connection will hang and keep listening to the consul until the state of the corresponding service changes. Thus, when other lookup threads (different services) send requests to consul, the request will be placed into the queue until the first service state change or timeout. 

And @BalloonWen and I think the correct performance should be that each lookup thread should have its corresponding connection instead of sharing a connection. By doing this, one lookup hang would not influence others.

stevehu:
The consul client was designed for HTTP/2 only as the HTTP/2 connection is multiplex. You need a connection pool if using HTTP/1.1. I think the first thing we need to find out why HTTP/2 stops working.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Consul connection freezing bug](https://trello.com/c/nrIxOVm6/247-consul-connection-freezing-bug)

PR #570, opened 2019-07-04
Fix/#523 consul http https connection

BalloonWen:
related issue: https://github.com/networknt/light-4j/issues/523
the previous fix didn't cover all the scenarios,
the previous fix of this issue will cause a problem:
- when a service itself is not registered in Consul, but trying to discover other services, it will create too many connections


stevehu:
@BalloonWen As we have discussed, I think we should cross-check the consulUrl to make the decision if HTTP/1.1 or HTTP/2 is used. If the consulUrl is http:// then HTTP1.1 should be used even the http2 is enabled in the consul.yml file. There are some users who do not know that Consul only supports HTTP/2 in the case HTTP/S is used. Thanks.

stevehu:
[Fix/#523 consul http https connection](https://trello.com/c/R08hULGQ/320-fix-523-consul-http-https-connection)

stevehu:
[Fix/#523 consul http https connection](https://trello.com/c/o37tJigx/321-fix-523-consul-http-https-connection)

Issue #518, opened 2019-05-23
NullPointerException when doing serviceToUrl


jiachen1120:
When a service can't be found in the serviceToUrl method, a NullPointerException is thrown. Not sure if this is the best thing to do, as it would be more expected to simply return null. 
 
In cases where a connection couldn't be made or something like that, a different exception would make sense.

stevehu:
An NPE is definitely not a good idea. We should log a ClientException with meaningful context message like with service cannot find the instance. 

One thing that we need to think about is the retry logic as the target service might be in a process to restart by the orchestrator like Kubernetes. We can also implement the capability to define a dummy return for each non-critical unavailable services.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [NullPointerException when doing serviceToUrl](https://trello.com/c/fqxLQAbT/317-nullpointerexception-when-doing-servicetourl)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [e89c5befcea144b4b00c53ecc38847809a99250d](https://trello.com/c/NumTpyMw/318-e89c5befcea144b4b00c53ecc38847809a99250d)

PR #576, opened 2019-07-11
Fix/npe service to url

jiachen1120:
issue: https://github.com/networknt/light-4j/issues/518

A quick fix for LightCluster.serviceToUrl method, return null instead of NPE when the corresponding service cannot be found.


stevehu:
@ddobrin The change to eliminate NPE from the framework. If there is no URL, then return null. In the application code, you need to make sure to check null. The other option is to throw an exception; however, this changes the signature of the method. Please advise which approach is the best for you. Thanks.

ddobrin:
I can make the necessary changes at my client and handle a null response when this change makes it into a release

It is more elegant than the exception.

stevehu:
Both 1.6.x and 2.0.x are merged with this PR. Thanks.

Issue #573, opened 2019-07-09
ConsulRegistry will not be notified when pods go down to 0


BalloonWen:
Consul registry will be notified when services that interested are changed, but will not be notified when pods go down to 0

stevehu:
Does it mean the last pod shutdown won't unregister from the consul? Is this a bug?

BalloonWen:
> Does it mean the last pod shutdown won't unregister from the consul? Is this a bug?

the last pod shutdown won't update light-4j ConsulRegistry, it's a light-4j bug.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [ConsulRegistry will not be notified when pods go down to 0](https://trello.com/c/Jw7tWDQC/327-consulregistry-will-not-be-notified-when-pods-go-down-to-0)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [ConsulRegistry will not be notified when pods go down to 0](https://trello.com/c/3wuPzU8T/328-consulregistry-will-not-be-notified-when-pods-go-down-to-0)

PR #577, opened 2019-07-12
Fix/#573 consul registry notify

BalloonWen:
related to issue: https://github.com/networknt/light-4j/issues/573
the subscriber will ignore the URL, and the listener will include the service which should be notified.
tested with Consul and direct registry for the notification part.


cluster/src/main/java/com/networknt/cluster/LightCluster.java
```diff
@@ -129,10 +130,19 @@ private URI toUri(URL url) {
...
+
+        ClusterNotifyListener(String serviceId) {
+            this.serviceId = serviceId;
+        }
+
         @Override
         public void notify(URL registryUrl, List<URL> urls) {
-            if(logger.isDebugEnabled()) logger.debug("notify is called in ClusterNotifyListener registryUrl = " + registryUrl + " urls = " + urls);
-            if(urls != null && urls.size() > 0) serviceMap.put(urls.get(0).getPath(), urls);
+            logger.debug("registryUrl is: {}", registryUrl);
```
BalloonWen: registryUrl is redundant, but to avoid changing the interface at this time, keep it for now.

stevehu: I think it is a good idea.

stevehu:
merged into both 1.6.x and master. Thanks

PR #578, opened 2019-07-12
randomly pick up port number for dynamic registry

stevehu:


server/src/main/java/com/networknt/server/Server.java
```diff
@@ -262,8 +269,8 @@ static private boolean bind(HttpHandler handler, int port) {
             server.start();
             System.out.println("HOST IP " + System.getenv(STATUS_HOST_IP));
         } catch (Exception e) {
-            if (!serverConfig.dynamicPort || (serverConfig.dynamicPort && serverConfig.maxPort == port + 1)) {
-                String triedPortsMessage = serverConfig.dynamicPort ? serverConfig.minPort + " to: " + (serverConfig.maxPort - 1) : port + "";
+            if (!serverConfig.dynamicPort || bindCounter > (serverConfig.maxPort - serverConfig.maxPort)) {
```
jiachen1120: Hi! Is this a typo? I think it should be `serverConfig.maxPort - serverConfig.minPort`

stevehu: It was my mistake. Thanks a lot for pointing it out.

stevehu:
With this PR, the behavior changed a little bit. Previously, the max port is never been used and now it is been used. Please let me know if you guys see any negative impact. Thanks.

server/src/main/java/com/networknt/server/Server.java
```diff
@@ -173,8 +175,13 @@ static public void start() {
...
                 throw new RuntimeException(errMessage);
             }          
-            for (int i = serverConfig.minPort; i < serverConfig.maxPort; i++) {
-                boolean b = bind(gracefulShutdownHandler, i);
+            while(true) {
+                bindCounter++;
+                if(bindCounter > (serverConfig.maxPort - serverConfig.minPort)) {
+                    break;
+                }
+                int randomPort = ThreadLocalRandom.current().nextInt(serverConfig.minPort, serverConfig.maxPort + 1);
```
jiachen1120: After research on TreadLoadRandom, if serverConfig.minPort >= serverConfig.maxPort + 1, an IllegalArgumentException would be throw. Should we catch and handle this exception?

stevehu: I think it is a good idea. We shouldn't allow the user to start the server in this case. I will make a change and test it to ensure the server stops. Thanks.

stevehu: There is a check before that. 

```
            if (serverConfig.minPort > serverConfig.maxPort) {
                String errMessage = "No ports available to bind to - the minPort is larger than the maxPort in server.yml";
                System.out.println(errMessage);
                logger.error(errMessage);
                throw new RuntimeException(errMessage);
            }          

```

jiachen1120: Oh, sorry, I didn't notice that

PR #580, opened 2019-07-12
fixes #579 when all instances are down and restarted, the client disc…

### LINKED ISSUE #579: when all instances are down and restarted, the client discovery might not triggered
The following is the sequence:

A consumer service calling a provider service and all providers are down. 
The serviceToUrl returns null in the consumer service. 
Consume service keeps retrying...
Provider service is up with one or more instances. 
The consumer service might not discover the newly started provider service instances. 

We need to remove the cache in the service map before return null in the second step above to force the discovery in the last step.

stevehu:
…overy might not triggered


stevehu:
The Server.Java update was a side effect in the previous cherry-pick merge.

stevehu:
@ddobrin I had a chat with @BalloonWen and it looks like we don't need to remove the empty array list from the service map anymore as we are checking the empty map in the condition check. What do you think?

PR #582, opened 2019-07-12
fixes #578 to use a hashset to track used ports

stevehu:


server/src/main/java/com/networknt/server/Server.java
```diff
@@ -269,7 +273,7 @@ static private boolean bind(HttpHandler handler, int port) {
             server.start();
             System.out.println("HOST IP " + System.getenv(STATUS_HOST_IP));
         } catch (Exception e) {
-            if (!serverConfig.dynamicPort || bindCounter > (serverConfig.maxPort - serverConfig.minPort)) {
+            if (!serverConfig.dynamicPort || usedPorts.size() >= (serverConfig.maxPort - serverConfig.minPort)) {
```
jiachen1120: usedPorts.size() >= (serverConfig.maxPort - serverConfig.minPort) + 1 would be more accurate？Since the final size of usedPort should equal to the capacity.

stevehu: @jiachen1120 I tried your code with +1 and the server won't quit when running out of port. I am using min 2000 and max 2003 for testing on my local with the service discovery consul folder. Let me debug into it. Thanks.

stevehu: When usedPorts is 3, you have tried 4 ports already, the last try is not reflected in the hashset yet.

stevehu:
[fixes #578 to use a hashset to track used ports](https://trello.com/c/q4n0Y3x9/333-fixes-578-to-use-a-hashset-to-track-used-ports)

stevehu:
[fixes #578 to use a hashset to track used ports](https://trello.com/c/GRnP1b5w/334-fixes-578-to-use-a-hashset-to-track-used-ports)

PR #587, opened 2019-07-19
fixes #586 FormData.FormValue cannot be handled by Jackson JSON parser

### LINKED ISSUE #586: FormData.FormValue cannot be handled by Jackson JSON parser
Recently, we have added a form parser in the BodyHandler to allow the schema validator to work with the form request from a single page application. In testing, we have found that the map object of the body contains the FormData.FormValue which cannot be handled by a JSON parser. Further, the map contains an array as the value even there is only a single value in the form. 

We need to make the following changes in order to allow the request body validator to work. 

1. only put the string value into the map instead of  FormValue. This can be done by calling FormValue.getValue()

2. handle the value differently based on the size. If there are more than one values for the key, use an ArrayList, otherwise, use String instead. 

This will ensure that a string value is available as form values for most of the case and the scheme can be defined as such.

stevehu:


body/src/main/java/com/networknt/body/BodyConverter.java
```diff
@@ -20,15 +20,26 @@
...
+ *
+ */
 public class BodyConverter {
     static Map<String, Object> convert(FormData data) {
         Map<String, Object> map = new HashMap<>();
         for (String key : data) {
-            List<Object> list = new ArrayList<>();
-            for (FormData.FormValue value : data.get(key)) {
-                list.add(value);
+            if(data.get(key).size() == 1) {
```
jefperito: miss space between "if" and "(data..."

body/src/main/java/com/networknt/body/BodyConverter.java
```diff
@@ -20,15 +20,26 @@
...
+                String value = data.getFirst(key).getValue();
+                map.put(key, value);
+            } else if(data.get(key).size() > 1){
```
jefperito: miss space between "if" and "(data..."

jefperito: miss space between and "> 1)" and "{"

body/src/main/java/com/networknt/body/BodyConverter.java
stevehu: Cool. The styled format can greatly improve the readability.

stevehu:
In most of the cases, we don't need the comment like this; however, this bug was escaped from our code review before due to the special data structure and interaction with the Jackson. I think the comment will help us to pay more attention to the detail in other cases. Also, it lets other developers know why we have made the change.

body/src/main/java/com/networknt/body/BodyConverter.java
```diff
@@ -20,15 +20,25 @@
...
+ * For the same form key, there might be multiple form values, we handle it differently based on the size.
+ */
 public class BodyConverter {
     static Map<String, Object> convert(FormData data) {
         Map<String, Object> map = new HashMap<>();
         for (String key : data) {
-            List<Object> list = new ArrayList<>();
-            for (FormData.FormValue value : data.get(key)) {
-                list.add(value);
+            if (data.get(key).size() == 1) {
```
BalloonWen: looks good.
data.get(key) won't be null even send only a key without value based on @jiachen1120 testing.

stevehu: Thanks for the test confirmation.

Issue #588, opened 2019-07-23
Health check handler cannot return result in Json format


jsu216:
When enabled, the result from HealthGetHandler is a string "OK" (without double quotes). The "Content-Type" in response handler is also empty.
This makes light4j based service cannot be health checked if deployed behind a ESB like "Data Power". Because the response message must be in a unique format, otherwise, an error like "Invalid Json data format" will be returned from Data Power.
We need an option in config, health.yml to return a Json format result.
This option must be turned off by default to keep backward compatible.
There is a RFC submitted and approved: https://github.com/networknt/light-rfcs/blob/master/light-4j/0023-Support-Json-Format-Response-In-HealthGetHandler.md

stevehu:
@jsu216 Should we return double-quoted OK or a full-blown JSON like {"result": "OK}? Depending on the client implementation, double-quoted OK might not be considered as valid JSON for some systems.

jsu216:
Yes, it is a good idea to return {"result": "OK"}.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Health check handler cannot return result in Json format](https://trello.com/c/JQUhbanj/346-health-check-handler-cannot-return-result-in-json-format)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Health check handler cannot return result in Json format](https://trello.com/c/lGHKgozC/347-health-check-handler-cannot-return-result-in-json-format)

PR #589, opened 2019-07-24
API-189: Return Json format result from HealthGetHandler

jsu216:
Issue #: https://github.com/networknt/light-4j/issues/588


jsu216:
I am not sure which branch it should be merged to. I set it to 1.6.x for now.

stevehu:
It will be merged into both 1.6.x and 2.0.x.

Issue #595, opened 2019-07-29
reset connection in Http2Client after certain number of requests


stevehu:
HTTP/2 supports multiplex but it has a limitation. After a number of requests, the same connection can not be used anymore. We have handled it outside of the Http2Client so far but more and more users forget to reset the connection. Let's investigate if we can do it in the Http2Client and set a default value for the number of requests per connection.

We have an example in the consul client implementation and it reset connection after 1 million requests by default.

stevehu:
This should be considered along with the connection pool implementation. https://github.com/networknt/light-4j/issues/572

Issue #572, opened 2019-07-09
implement a connection pool for Http2Client in case the server is HTTP/1.1


stevehu:
HTTP/2 support multiplex and we can use one connection to send multiple concurrent requests. However, if the server doesn't support HTTP/2, then the connection will be downgraded to the HTTP/1.1 and we need a connection pool to allocate one connection per request.

jiachen1120:
@stevehu 
Through the investigation, I found that whenever the method `getRequestService()` inside http2client is called, regardless of whether  enable http2 or not, a new connection will be established. The connection type will be http2ClientConnection or httpClientConnection, depending on whether the target server supports http2. So for the current http2client, the type of connection does not affect the functionality.

I think maybe we should build a connection pool for http2connection based on host instead of create one for http/1.1. This avoids creating multiple http2connections for the same host. And for httpConnection. Each request to establish a new connection may be able to avoid blocking the connection as much as possible. (I think the scenario is different from consul)

ddobrin:
At my client, we had this issue and it was a showstopper.

We introduced an (older) REST client and used that one to connect to the downstream service.

jiachen1120:
@ddobrin Hello, Dan! Could you please specify the use cases of http2client for that showstopper? Have you use the `getRequestService()` method or using `connection.sendRequest()` to call the downstream service? That would be very helpful! Thanks!

stevehu:
@jiachen1120 getRequestService is a new API added for the circuit breaker and completable future. You are right that it creates a new connection every time it is called. I think we can refactor it to share the same connection if possible. @chenyan71 @yibogithub @jefperito What do you think?

I like the idea to associate the connection pool with the host regardless of HTTP/1.1 or HTTP/2. Once the connection is created, we can create multiple instances (HTTP/1.1) or single instance (HTTP/2). 

@ddobrin I remember that you mentioned that by specifying http2 does not guarantee the connection created is HTTP/2 and you have a workaround. Could you please elaborate the details?

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [implement a connection pool for Http2Client in case the server is HTTP/1.1](https://trello.com/c/tXzEEDM9/402-implement-a-connection-pool-for-http2client-in-case-the-server-is-http-11)

stevehu:
[...]

Issue #11, opened 2017-01-07
Find the best location to inject server info to the routing handler


stevehu:
Light-Java is designed to work with swagger specification to generate the code and trigger scope verification in security and validation in validator. However, this make injection of common services with is not defined in the service swagger specification hard to be injected and by pass all the swagger validation. Need to find a way to inject server info which is the endpoint that output all information about components wired in and their configuration.

AmaranthLIS:
Yes, if project generated by swagger we can't add other handlerProviders like info/healthy ...
Even if add new handlerProvider to com.networknt.server.HandlerProvider
But it is so need!

stevehu:
Two months ago, I have tried to inject the a separate path with /server/info and a specific scope to control who can access it in SwaggerHandler. It is working but I just don't feel good about it. One problem with this is the final url is dynamic as the API/service swagger spec might have v1 or v2 as base path and it is not predictable. We need a static server info url so that it can be called from portal. I am working on the service registry and discovery now and hoping to let the server to send notification to portal once it is up and running with full path of server info url just like it register itself with Consul or Zookeeper. The other way to work our is inject a handler into the call chain to bypass swagger, validation etc. It is doable but pretty messy as call chain is dynamic per API/service. I am open to any ideas or recommendations at the moment and will come back to this issue soon as this is needed for production like you said.

AmaranthLIS:
In this cause i really don't know how better it do...
I can offer this way: add PATH an exception for swagger-validation which get from config. 
HandlerProvider is manually added routing (for example /server/info).

stevehu:
This is the approach I was taking but due to the dependency of base path in the original swagger spec, the server/info might have /v1 or /v2 or some other prefixes. Also, you don't need to by pass the validation and security as when you inject something the scope and other validation constraints will be in. I am hoping I will have answers once service registry is completed and see if we can let server to register the server/info endpoint on consul or zookeeper.

stevehu:
Tried to convert ServerInfoHandler to middleware so that it can be put into the middleware chain. It works but it doesn't make sense that every request must go through ServerInfoHandler to check it the uri is matched. It is too much overhead just for this simple common API.

stevehu:
The next step is to secure the endpoint so that only light-portal can access it.

stevehu:
[...]

PR #598, opened 2019-08-01
Feat/connection pool

jiachen1120:
Related issue: https://github.com/networknt/light-4j/issues/572 https://github.com/networknt/light-4j/issues/595

This PR implements the connection pool of http2client to optimize the connection establishment of the http2client module.

**implementation**

By embedding the connection pool into an existing method
```java
CompletableFuture<ClientResponse> callService(URI uri, ClientRequest request, Optional<String> requestBody)
```
, the user can call this method or `getRequestService(URI uri, ClientRequest request, Optional<String> requestBody)` which combine the `callService()` with circuit breaker to cache the established connections and reuse them without considering the type of connection.

When the connection is http/1.1, multiple connections will be added to the connection pool for the same host. The workflow is to monitor whether there are connections that are not yet occupied by other requests and are not waiting for a response when a request is sent. If so, reuse the connection until the connection is closed or the maximum number of requests is reached, otherwise a new connection is established and cached into connection pool for this host.

When the connection is http/2, only one connection will be added to the connection pool for the same host. When sending a request, reuse it until the connection is closed or the maximum number of requests is reached. When the connection is not encrypted(using the http protocol), more than one connection will be added to the connection pool. This is because the undertow will first establish a http/1.1 connection and then upgrade it to http2 instead of directly establishing http/2 connections.

The connection pool uses the LRU mechanism. Least recently used connection will be removed when the connection pool size reaches the set value. 

**Configuration** 

Users can configure the size of the connection pool; the maximum number of requests per connection and whether http/2 is enabled by configuring client.yml.

Or, user can clean up the connection pool manually by calling `Http2ClientConnnectionPool.getInstance().clear()`

For example,
```yml
request:
  errorThreshold: 2
  timeout: 3000
  resetTimeout: 7000
  enableHttp2: false
  connectionPoolSize: 1000
  maxReqPerConn: 1000000
```

**Example**

```java
try {
    ClientRequest requestB = new ClientRequest().setMethod(Methods.GET).setPath(apibPath);
    
    CompletableFuture<ClientResponse> response = client.callService(new URI(apibHost), requestB, Optional.empty());
    
    int statusCodeB = response.join().getResponseCode();
    if (statusCodeB >= 300) {
        throw new Exception("Failed to call API B: " + statusCodeB);
    }
    String apibList = response.join().getAttachment(Http2Client.RESPONSE_BODY);
} catch (Exception e) {
    logger.error("Exception:", e);
    throw new ClientException(e);
}
```


stevehu:
@jiachen1120 This is so cool and it is a long-waited feature for a lot of users. Thanks a lot for putting it all together. I haven't read the code yet and I might be wrong. Regarding the HTTP/1.1 connection reuse, my understanding is that one connection can only be reused in the same thread as it would be too hard to sync different thread than picking another connection from the pool. I need to understand the implementation first but something comes up when I read your document above and I don't want to forget about it.

jiachen1120:
@stevehu Hi! Thanks for quick response! 

In order to deal with the above situation. I created a state map to record the usage of each connection. For http/1.1, there are three states: `hanging`, `available`, `close`. Whenever a request is sent, a connection with the status `available` will be randomly picked from the connection pool, and immediately change its state to `hanging` to prevent other thread gets the connection. This method is synchronized, so that at http/1.1, different threads would not get the same connection with the available status.

When a connection response returns, the method responsible for resetting the status is called, and the status of connection is reset to `avaliable` if the connection is not closed or the maximum number of requests is reached.

After testing, it seems work properly. I will further considering about it.

stevehu:
The logic sounds very good. For HTTP/1.1 connection, do we need to limit the number of the request per connection? I think the number of multiplex requests is the limitation for HTTP/2 only. Thanks.

jiachen1120:
@stevehu 
Yes, I think you are right. I can remove this check for http/1.1 if there is no problem with unlimited reuse of http/1.1.

I have not tested whether there are any restrictions on the reuse of http / 1.1. Because it is not easy to test, I also did not find the ideal answer on google. So I chose a more insurance method, which is a unified restriction with http 2. 😂

stevehu:
Yes. There is no negative impact to set a limit just like the HTTP/2. The limit was introduced in HTTP/2 as the designer using a long number to track the frames. It is considered a bug by a lot of people and will be removed in the future of the protocol.

ddobrin:
@jiachen, @stevehu, 
please allow me to make some notes on this PR.

**1. Reset by request count or time**
For HTTP2, we discover the host using the cluster class:
// discover the host
host = cluster.serviceToUrl("https", serviceID, TAG, null);

we use this host to connect() or to pass into this new method callService():
connection = client.connect(new URI(host), Http2Client.WORKER, Http2Client.SSL, Http2Client.BUFFER_POOL,OptionMap.create(UndertowOptions.ENABLE_HTTP2, useHttp2)).get();

or 
client.callService(new URI(host), request, ....)

The callService() will get a connection from the pool and, if it reaches one of the 2 limits, resets the connection, but within the same URI ! This solves only partially the reset problem, but doesn't balance across multiple hosts, if discovered . from Consul for example.

**we have 2 choices:**
a.  add another class to discover/re-discover, keep count, and set a new config values. We could also leave that to API Teams to build
b. the functionality still has to be built, however we expose the reqcount or time limit as getters in the HTTPClient, if we wish to reuse the same values.

I must implement the first one for my client and it can be shared if found useful.

**2. When a callService call is made** by a microservice, and a different host is passed, and it is HTTP2, the new connection is being added to the pool, however the previously used one is **not evicted from the pool** 

**3. HTTP1.1 - where we have multiple connections in the pool.**

If we set only the maximum size, and the pool is used up (say limit=10), while the microservice is under load, we create a new connection past the limit, however we return it to the pool and evict the oldest.
If the service stays under load, we will in fact create a new connection for every call, which defeats the purpose of the pool.

Can we consider a min and max (HTTP1.1), where we let the pool grow under load (assume a maximum acceptable max size), however let it shrink when not under load.

For example, if the load capacity is dropping at 60-70%, shrink the pool.

**4. The reset count/reset by time needs to be added for HTTP1.1 as well**, as we'd like consumers to reestablish those connections as well.
Please read this comment in line with [1], as we could leave the "rebalancing" up to API developers or not

I think the ideas are all good, maybe we can fine tune/add a bit to address [1-4] above

stevehu:
@ddobrin Thanks a lot for the long comment. I agree that we need to re-balance the backend cluster for each connection reset. I missed this important feature and mistakenly asked @jiachen1120 to remove the reset for HTTP/1.1 connection. 

For point 1, I think we should implement it in the client module instead of asking our users to take responsibility unless there is no easy way to do so.

For point 2, we are using the uri as the key to look up the connection in the pool. So for the same uri, there is only one connection in the pool. I don't think we should remove other connection when a new uri is introduced. Chances are the old uri will be used by another call and the existing connection will be retrieved to complete the request. Is there any reason that we need to evict an HTTP/2 connection?

For point 3, I think the maximum/minimum pool size is a good idea if it is not too hard to implement. This gives users a lot of flexibility and makes the connection pool more adaptive to burst of loads. I think most JDBC connection pool is doing the same. 

For point 4, it is something I missed. The connection reset is not just to fix the HTTP/2 bug but to re-balance the backend service instances as we are using client-side discovery. @jiachen1120 I am sorry to ask you to remove the reset logic.

ddobrin:
[1]
I agree with you @stevehu about the best solution is implementing it in the client, this is what was discussed in the past. 
I am planning for the case that it can't be added easily, then it's a feature which clients will have to handle, and I am OK with that. 

We can do that if we add an additional method (better said, a separate one), or modify this one, where you can pass it serviceID and Tag, then deal with the call within that method.

I am OK with the current control I have with clustering, connection and invocation in a utility class, as I have different security models for my APIs.

An additional class/method should allow consumers to have that flexibility.

And, let's not forget, different  users have different requirements, while several APIs are already deployed in the Cloud and change is not instantaneous.

[2] that connection might be stale, and we want the calls to balance to the next URI; that's the rationale

[3] The min and max is required, I would say, for the following reason:
right now, say max is set to 10, but request #11 arrives. 
pool is max'ed out
a new connection is created
pool stays at size 10
the new connection is added to the pool and the oldest evicted.

If the pool is set to low, or the influx of requests is high, each new requests results in creating a new connection. 

This is why I am suggesting it.

My most important concern is [1]. I can help out if needed

stevehu:
@ddobrin Thanks for the detailed explanation which brings us on the same page. For point 2, it raises another question about the design of our connection pool. What is the key to the connection pool? serviceId or URI? If we use serviceId, it is before the service discovery and we can keep all the connections to each backend service per serviceId. If the key is URI, then the new connection should only be created when a connection is reset or re-balanced. At that time, we don't know how to remove the previous connection. The best place to remove the previous connection is when the limit is reached I think. This is pretty complicated and I can set up a zoom meeting if you guys want to have a group discussion.

ddobrin:
We can discuss it over Zoom, if you guys wish, however my time is limited to tomorrow morning

jiachen1120:
@ddobrin @stevehu 
Thanks for the reviews and suggestions!

[1] [4] Rebalancing is a good point, sorry I ignored this in my previous design. I will find a way to integrate this method in the client. I will delve further into whether introduce a new method or make changes on existing methods.
[2][3] Currently, The connection pool I design can be showed as the following chart:
```
| uris           | connections   list                                |
|----------------|---------------------------------------------------|
| localhost:8080 | Http2ClientConnection                             |
| localhost:8081 | HttpClientConnection1, HttpClientConnection2, ... |
| ...            |                                                   |
```
The key is URI, each uri will corresponding to one or multiple connections depend on the connection type(http/2 or http/1.1)
Oldest entry(URI->connections) instead of oldest connection will be evicted from the connection pool when the **URI** number reach the limitation of the pool size. Therefore, I think maybe 
`If the service stays under load, we will in fact create a new connection for every call, which defeats the purpose of the pool` won't happen in this case.
I still think the maximum/minimum is good idea, but maybe not for the entire connection pool. Instead, it is for each entry. To be more specific, for the connection list of each entry.

Now I just evict a single connection by checking if the connection is closed or reach the maxReqCount. If using the maximum/minimum method, It would be better remove the outdated connections.

ddobrin:
@jiachen1120 : here's the snippet which I use in a utility and I think that this is what we can incorporate. 

`			try {
				// discover the host and establish a connection if not established at start-up
				host = cluster.serviceToUrl("https", serviceID, TAG, null);
				// if discovery has failed, log the error and throw a CLientException
				if(host == null) {
					logger.error("Failed to discover service with serviceID: {}, and tag: {}", serviceID, TAG);
					throw new ClientException(String.format("Failed to discover service with serviceID: %s, and tag: %s", serviceID, TAG));
				}
				
				connection = client.connect(new URI(host), Http2Client.WORKER, Http2Client.SSL, Http2Client.BUFFER_POOL,
						OptionMap.create(UndertowOptions.ENABLE_HTTP2, useHttp2)).get();
			} catch (Exception e) {
				logger.error("Failed to connect to host: " + host, e);
				throw new ClientException(e);
			}`

jiachen1120:
@ddobrin Thanks for sharing!!👍🏻 
yes, I agree. We should not rush for this feature. We need to make sure it is reliable and does solve our existing problems.

chenyan71:
I am not clear about the issues. But maybe we should take a look the test cases we created on light-consumer-4j. We have concurrent service call support already on Https2Client which we create new http connection for every API call. And I don't think create new http request connection is big cost.

stevehu:
@chenyan71 Creating a new connection per request won't have the reset issue. There are two options to create a new connection: use the same host discovered before or discover a new host. For option 1, it takes time for the handshake and establishes the connection but it is not too bad. For option 2, it resolves the re-balance issue but putting a lot of pressure on the Consul nodes or Consul agent and it would be very slow. Either HTTP/1.1 or HTTP/2 will greatly benefit from reusing an existing connection. The connection pool implementation is just a natural extension and makes the reuse easier. Of course, it increases the complexity but with proper tests, I think it is manageable.

chenyan71:
@stevehu Agree. The connection pool will helpful for sending request multiple times to same API endpoint. So make it optional and configurable will be better option.

ddobrin:
@jiachen1120 @chenyan71 @stevehu : 
I think configuration should allow you to decide whether you wish to re-establish a conn with the same host or re-discover.
This can be achieved also by exposing different methods.

The values for reseting (time/no of req) can be the same in both cases.

I need to use the re-connect and re-discover not just for load balancing, but to address also a Consul related issue (details left out here).

@chenyan71 : you will have the same issue when the APIs at your client site will go live 👍

jiachen1120:
@stevehu @ddobrin Hi!
In the latest comments I have added the following improvements:
1. Provide an overloading method for callservice() to embedded service discovery and load balancing
2. Add the minConnectionNumPerHost/maxConnectionNumPerHost configuration in client.yml to control the number of connections for each host. When the number of connections exceeds 75% of the maximum number of connections, the number of connections will shrink to the minimum number of connections.
3. Add connectionExpireTime configuration in client.yml. Therefore, a connection will be evicted if: (1)connection closed (2)reach the maximum request limitation (3) expired

stevehu:
@jiachen1120 I see that the newly added callservice will always perform service discovery. Is this method only be called when you want to reset the connection?

jiachen1120:
@stevehu Hi! The newly added one will always perform service discovery (cached after the first time, the local cache will be notified when consul status change). Therefore, there will be a balance between each host. 

The connection will be reset for both callServices method (call service by using uri directly/ call service by using service Id) in three scenarios: (1)connection closed (2)reach the maximum request limitation (3) expired.

PR #604, opened 2019-08-19
fixes #603 add JwtVerifier to replace JwtHelper

### LINKED ISSUE #603: Using JwtHelper on restful API for JWT security token config file issue
Based on new design for the restful API, system will use openapi-security.yml as JWT security token configuration file. And new version light-codegen does not generate security.yml file for restful API anymore. 

In JwtVerifyHandler class, system check the openapi-security.yml  first, if not find, use security.yml for backward compatibility:

  // check if openapi-security.yml exist
  config = Config.getInstance().getJsonMapConfig(OPENAPI_SECURITY_CONFIG);
  // fallback to generic security.yml
 if(config == null) config = Config.getInstance().getJsonMapConfig(JwtHelper.SECURITY_CONFIG);

But in JwtHelper class, it only load security.yml for JWT config:

    public static final String SECURITY_CONFIG = "security";
    static Map<String, Object> securityConfig = (Map)Config.getInstance().getJsonMapConfig(SECURITY_CONFIG);
    static Map<String, Object> securityJwtConfig = (Map)securityConfig.get(JWT_CONFIG);

I think it should have same logic to check openapi-security.yml since resful API will use the file for configuration.

stevehu:


stevehu:
Tested with the petstore application in light-example-4j and everything works fine.

info/src/main/java/com/networknt/info/ServerInfoGetHandler.java
```diff
@@ -89,8 +86,20 @@ public void handleRequest(final HttpServerExchange exchange) throws Exception {
...
-            secMap.put("oauth2FingerPrints", JwtHelper.getFingerPrints());
+        // as we have replaced the static JwtHelper to JwtVerifier, we cannot use the static method to
+        // get the fingerprints anymore, need to iterate the registered security module to do so.
+        Map<String, Object> moduleMap = ModuleRegistry.getRegistry();
+        // use set to eliminate duplications and will convert to ArrayList later
+        Set<String> fingerprints = new HashSet<>();
+        for(Object module: moduleMap.entrySet()) {
+            if(module instanceof IJwtVerifyHandler) {
+                IJwtVerifyHandler handler = (IJwtVerifyHandler)module;
+                fingerprints.addAll(handler.getJwtVerifier().getFingerPrints());
```
BalloonWen: @stevehu , I have a question, I think the purpose is: comparing to getting static JwtVerifier, now it can choose different JwtVerifier to get different fingerprints. But why it has to implement the IJwtVerifyHandler to get the JwtVerifier instead of getting from the service.yml?

BalloonWen: In this case, the handler doesn't have to implement IJwtVerifyHandler but inject different JwtVerifier in service.yml. And I think this is the purpose of this piece of code.

stevehu: @BalloonWen Good question. As users can define the handler chain from both service.yml and handler.yml, it is not reliable to use only one of them. The only reliable source is the registered module map as this map contains only the enabled middleware handlers. I don't want the info module to depending on any JwtVerifyHandler from each framework, so an interface will remove that dependency so that I can invoke the interface method to get the JwtVerifier from all the injected JwtVerifyHandlers.

BalloonWen: Hi Steve, I'm not referring to define the handler chain in service.yml but to define JwtVerifier in service.yml. in this case, it's not depending on JwtVerifyHandler either, and more straightforward, because eventually, you want to get the JwtVerifier, either you put the JwtVerifier in an IJwtVerifyHandler, and using the IJwtVerifyHandler.getJwtVerifier() to get it. Or, you can just define the JwtVerifier in service.yml directly?

stevehu: I understand now. Yes. That is another option and we can generate service.yml for each framework. However, it adds the dependency for service.yml or at least an entry in service.yml. When users are using light-codegen, it is OK. However, if they are starting a new project from scratch or their projects have two or more frameworks combined (REST and GraphQL), then it is very hard for users to figure out how to make the configuration as there might be two JwtVerifier instances for each framework corresponding to openapi-security.yml and graphql-security.yml. As a framework developer, my thinking is always to make our user's life easier if possible. In the current case, all classes involved are part of the frameworks and we already have the classes in the registered module. The current drawback is obvious, if someone is trying to extend the framework and provide another JwtVerifyHandler, he/she needs to remember to implement the interface.

BalloonWen: That makes sense. Thank you for the comments. :)

Issue #625, opened 2019-11-29
Add ByteBuffer clientCallback support at Http2Client on client module


chenyan71:
On undertow HttpServerExchange (io.undertow.io.Sender)can send response by String or ByteBuffer. But light-4j client module Http2Client ClientCallback only can handler string by using UTF8Output. Which will cause some types of response can not be received by client call. For example PDF download.

We need add one more  ClientCallback to handler ByteBuffer in response.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Add ByteBuffer clientCallback support at Http2Client on client module 1.6.x](https://trello.com/c/ZJoXrolH/510-add-bytebuffer-clientcallback-support-at-http2client-on-client-module-16x)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Add ByteBuffer clientCallback support at Http2Client on client module master](https://trello.com/c/HAkMcTu2/511-add-bytebuffer-clientcallback-support-at-http2client-on-client-module-master)

PR #626, opened 2019-11-29
fixes issue #625

chenyan71:
add new callback for Http2Client to handler bybuffer response

### DIFF
--- a/client/src/integration/java/com/networknt/client/Http2ClientIT.java
+++ b/client/src/integration/java/com/networknt/client/Http2ClientIT.java
@@ -674,6 +674,29 @@ public class Http2ClientIT {
         return reference.get().getAttachment(Http2Client.RESPONSE_BODY);
     }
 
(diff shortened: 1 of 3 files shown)


client/src/integration/java/com/networknt/client/Http2ClientIT.java
```diff
@@ -674,6 +674,29 @@ public String callApiAsync() throws Exception {
         return reference.get().getAttachment(Http2Client.RESPONSE_BODY);
     }
 
+    public ByteBuffer callApiWithByteBuffer() throws Exception {
```
jiachen1120: Hi! After manual testing, the byte array response can be handled properly! But this test method seems not be invoked and no @test annotation. Does this need to be added?

chenyan71: I will take a look

Issue #629, opened 2019-12-13
make Http2Client.SSL private to block users to use it directly


miklish:
In light 1.6.11, the static variable Http2Client.SSL is no longer eagerly initialized. This is a functionally breaking change from light 1.6.10.

https://github.com/networknt/light-4j/pull/612/files#diff-905ed9c154545ecedeab300229fa9b99

Applications that assume the SSL variable is initialized, may fail mysteriously. I suggest that either:

1. Http2Client.SSL is deleted from Http2Client in light 1.6.12
2. Http2Client.SSL is deprecated

In either case, the changelog (and also possible deprecation comments) should contain documentation stating that SSL is no longer eagerly initialized, and that the Http2Client.getDefaultXnioSSL method should be used instead (https://github.com/networknt/light-4j/pull/624).

My team would personally like to see option 1. While it could cause build failures in user applications, we believe that this is an appropriate result, as it will force end users to examine their use of the SSL variable and save them having their code build, but then fail mysteriously when run.

stevehu:
@miklish It can be removed. For normal applications, they are using the connect method of Http2Client so nobody realized that the SSL is lazily initialized. In my mind, it won't break existing applications. The only application that is impacted is the light-router which is using the Http2Client.SSL for the Undertow Client used in the router. It was a hack in my mind and the risk is controlled.

miklish:
Ok cool. I will create a PR to make it private. Thanks

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [make Http2Client.SSL private to block users to use it directly](https://trello.com/c/QTI0tp3t/508-make-http2clientssl-private-to-block-users-to-use-it-directly)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [make Http2Client.SSL private to block users to use it directly master](https://trello.com/c/ZgasTnTH/509-make-http2clientssl-private-to-block-users-to-use-it-directly-master)

PR #630, opened 2019-12-13
Make SSL private to prevent its usage before it is initialized. 

miklish:
See issue #629 for full details.

### DIFF
--- a/client/src/main/java/com/networknt/client/Http2Client.java
+++ b/client/src/main/java/com/networknt/client/Http2Client.java
@@ -83,7 +83,7 @@ public class Http2Client {
             .set(Options.KEEP_ALIVE, true)
             .set(Options.WORKER_NAME, "Client").getMap();
     public static XnioWorker WORKER;
-    public static XnioSsl SSL;
+    private static XnioSsl SSL;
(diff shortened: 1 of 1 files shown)


miklish:
Found another usage of SSL in light-4j (OAuthHelper)… WIll fix and update PR branch

miklish:
Sorry folks... as mentioned, there are more cases where the variable SSL is used directly that weren't updated... updating now

stevehu:
I think this breaks a lot of existing applications. If we don't do anything, all existing applications should work.

miklish:
I fixed it so it builds now, but the Audit Handler unit tests are failing. I'm examining why this is.  As for breaking existing applications, that is a good thing. The alternative is apps that build but fail when run.

miklish:
I figured out why some of the unit tests were failing. They use Http2Client.SSL in the test, but do not have any truststores configured. The solution is simply to delete the Http2Client.SSL argument.

The problem was that I had replaced the Http2Client.SSL argument with Http2Client.getInstance().getDefaultXnioSSL() which caused the client to look for truststores that were not configured.

miklish:
Ok, as per our discussion, this change should be reverted, as it causes to many issues in other related repos. We will instead deprecate Http2Client.SSL

Issue #632, opened 2019-12-19
ConsulRegistry keeps the first service url discovered in cache


jsu216:
# Brief
We found a bug in ConsulRegistry that the registry cache (in AbstractRegistry) keeps the first service url discovered, which is not sync with the Consul service registration changes happen after.

## Symptom
When a service (e.g. Consumer Proxy) subscribe a service through `ConsulRegistry`, it does get notified when new service instance becomes available or offline. If all service instances gets deregistered, however, `ConsulRegister` still remembers the first service url discovered. If any request comes in at this time, the consul client will "discover" that outdated service url as a result.
If any new service becomes available after, consul client can discovery the new service url, but will keep "discover" the original url when no service available again.
This behaviour can also be verified through the debug logs from `LightCluster` in this scenario. 

## Investigation
`ConsulRegistry` has its own cache that holds the result of previous service discoveries. The cache is maintained by its own notification listener which keep the discovery result sync with the changes in Consul registry.
`LightCluster`'s discover method call the `subscribe` and `discover` methods in the `Registry` interface when there is no service url found in its own cache (Line # 103-108).
The problem is the  implementation of the `subscribe` and `discover` method called by `LightCluster` are implemented in base class `AbstractRegistry`, which has its own cache for the discovery result (i.e. `subscribedCategoryResponses`). Unlike `ConsulRegistry`, this cache does not get updated when service registration changes in Consul.

## Solutions

1. In `LightCluster`, do not call `subscribe` or `discover` method from the `Registry` interface.
This seems to solve the issue, but also breaks some existing unit tests. I think the calling for discover method is required for other Registry implementation like DirectRegistry
2. When `LightCluster` subscribe the service, do not call the "notify" method which is implemented in the `AbstractRegistry` class. So that the unmanaged cache in `AbstractRegistry` will not be populated with any value and no outdated content will be stored as a result. This solution passes current unit tests.

I will create a PR using the second solution.

stevehu:
@jsu216 Nice findings and good analysis. In hindsight, we shouldn't make the service registry and discovery this complicated. The original idea is to support others like zookeeper and etcd, etc. But now, only Consul is being used widely.

stevehu:
@yibogithub Is this the same issue your project was faced?

ddobrin:
It is correct that the first URL in the discovered list is the one which was cached.
[...]

PR #633, opened 2019-12-19
Bug fix: ConsulRegistry keeps the first service url discovered in cache

jsu216:
Fix the [issue](https://github.com/networknt/light-4j/issues/632)


jsu216:
@stevehu I move the code change from `CommandFailbackRegistry` to `ConsulRegistry` to avoid any possible unexpected change to `ZooKeeperRegistry`.

Issue #649, opened 2020-02-15
Service discovery failure at beginning of service startup 


chenyan71:
Scenario:
when default consul config:  enableHttp2=false
In the service API, the service method need call multiple downstream service API.
At   beginning of service startup, some API Service discovery from consul by service Id will failed, and after for while, the become success with no error.

Code:
It looks like thread issue. in ConsulRegistry, for some serviceName:

      urls = serviceCache.get(serviceName);

The urls return empty list. It means the thread load the service with empty url value from consul discovery.

Changed the consul config to enableHttp2=true. The issue disappear.

chenyan71:
It happened only when use https connect to consul. It works for http consul connection

stevehu:
I think this is the same issue. https://github.com/networknt/light-4j/issues/634

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Service discovery failure at beginning of service startup 1.6.x](https://trello.com/c/DtZo6n9W/573-service-discovery-failure-at-beginning-of-service-startup-16x)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Service discovery failure at beginning of service startup master](https://trello.com/c/oZlAJfZW/574-service-discovery-failure-at-beginning-of-service-startup-master)

PR #652, opened 2020-02-25
 fixes for #649

stevehu:


stevehu:
@chenyan71 I have created the PR from your branch and merge the latest code to your branch. While reviewing the changes, I found that we cannot make the change in the ConsulClientImpl to use HTTP/2 connection when the https protocol or enableHttp2 is true. In case that enableHttp2 is not in the config file, even we use the Http2Connect but the request we send to the Consul will still be HTTP/1.1 which cannot leverage the multiplexing. What will happen if we rollback this change and only leave the ConsulRegistry change to check if the urls is empty? Thanks.

stevehu:
The isHttp2 is tested and confirmed by @chenyan71 and it is ready to merge.

PR #711, opened 2020-05-16
fixes #709

### LINKED ISSUE #709: JwtVerifier throw NullPointerException if user remove the jwtr/certificate from openapi-security.yml
In the openapi-security.yml (start from ligh-codegen), light-4j define default certificate for jwt token:

jwt:
  certificate:
    '100': primary.crt
    '101': secondary.crt

If the jwt token key is not in the config list (by default: 100/101), system will go to key server to retrieve certificate (client.yml). 

But in some case, user may not want to define default certificate and try to remove it from openapi-security.yml, then service API get error on service startup:

Caused by: java.lang.NullPointerException
         at com.networknt.security.JwtVerifier.<init>(JwtVerifier.java:120)
         at com.networknt.openapi.JwtVerifyHandler.<clinit>(JwtVerifyHandler.java:79)

chenyan71:


chenyan71:
OK, I will change the file name.
what about: openapi-security-no-default-jwtcertificate.yml

stevehu:
That would be good as long as it is not called security.yml and openapi-security.yml

PR #712, opened 2020-05-31
Update BodyConverter.java

thirtysixmm:
Bug fix when using File as the FormData value.

### DIFF
--- a/body/src/main/java/com/networknt/body/BodyConverter.java
+++ b/body/src/main/java/com/networknt/body/BodyConverter.java
@@ -26,20 +26,32 @@ import java.util.*;
  */
 public class BodyConverter {
     static Map<String, Object> convert(FormData data) {
-        Map<String, Object> map = new HashMap<>();
-        for (String key : data) {
-            if (data.get(key).size() == 1) {
-                String value = data.getFirst(key).getValue();
-                map.put(key, value);
-            } else if (data.get(key).size() > 1) {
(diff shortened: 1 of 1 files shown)


stevehu:
@thirtysixmm Thanks a lot for the enhancement. My understanding is that it is not introduced by the new version of Undertow but a missed feature in the original implementation. Please correct me if I am wrong.

stevehu:
[Update BodyConverter.java to handle files master](https://trello.com/c/TRoIuWRS/668-update-bodyconverterjava-to-handle-files-master)

thirtysixmm:
Thank you for accepting my PR. Not sure what changed. I used to read form data value like the below code before upgrade

```
Map<String, Object> body = (Map) exchange.getAttachment(BodyHandler.REQUEST_BODY);
ArrayList<FormData.FormValue> data = (ArrayList<FormData.FormValue>) body.get("data");
ArrayList<FormData.FormValue> files = (ArrayList<FormData.FormValue>) body.get("files");
```

After upgrading to the latest version, my handler code changed to.

```
Map<String, Object> body = (Map) exchange.getAttachment(BodyHandler.REQUEST_BODY);
String data = (String) body.get("data");
FileItem file = (FileItem) body.get("files");
```

stevehu:
It looks like the FileItem is newly introduced in the latest version. Thanks a lot for the investigation and fix.

Issue #715, opened 2020-06-04
Fix Consul connection conflicts when using HTTP/1.1


miklish:
When connecting to Consul using a non-HTTP/2 connection and 2 (or more) threads simultaneously need to get a url for the same service, they can potentially both end up in the `ConsulClientImpl.getConnection` method.

Refers to PR #716 

Method call chain:

```
LightCluster.serviceToUrl(serviceName)
--> LightCluster.discovery
--> CommandFallbackRegistry.discover (calls super)
--> AbstractRegistry.discover
--> CommandFallbackRegistry.doDiscover
--> ConsulRegistry.discoverService
--> ConsulRegistry.lookupServiceUpdate
--> ConsulRegistry.lookupConsulService
--> ConsulClientImpl.lookupHealthService
--> ConsulClientImpl.getConnection(String cacheKey)
```

Currently, for each `cacheKey`, the `getConnection` method will return the same `ConsulConnection` regardless of the thread that requests it. This can result in both threads using the same `ConsulConnection` simultaneously, generating a bad state in the shared `ConsulConnection`. In practice, this has resulted in NPE's in the `serviceToUrl` method, and has also resulted in the service being unable to make subsequent requests to Consul.

The solution is to ensure that every thread has its own private `ConsulConnection`. The downside of this is that it can cause a large cache of `ConsulConnection` objects (and hence HTTP connection objects) being created. However, the connection cache size cannot grow without bound, as it will be limited by the size of the undertow worker thread pool.

Additionally, with reasonable connection  timeouts in Consul and the user service, unused connections contained in cached ConsulConnection objects will eventually be closed, freeing up network resources. If a ConsulConnection object containing a closed connection is requested, a new active HTTP connection will created automatically.

The methods...

- checkPass
- checkFail
- registerService
- unregisterService
- lookupHealthService

...have this issue, as they all call getConnection, and don't use keys that are thread-specific.

NOTE: It is recommended to simply use HTTP/2 to connect to Consul, since a connection cache is not needed with HTTP/2.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix Consul connection conflicts when using HTTP/1.1 1.6.x](https://trello.com/c/H5D1vp7n/670-fix-consul-connection-conflicts-when-using-http-11-16x)

stevehu:
[...]

PR #716, opened 2020-06-04
fix for issue #715 - ensure cached consul connections are specific to…

miklish:
… a thread to avoid concurrent use of the connections across threads.

See issue #715 for details.

### DIFF
--- a/consul/src/main/java/com/networknt/consul/client/ConsulClientImpl.java
+++ b/consul/src/main/java/com/networknt/consul/client/ConsulClientImpl.java
@@ -104,7 +104,7 @@ public class ConsulClientImpl implements ConsulClient {
 		logger.trace("checkPass serviceId = {}", serviceId);
 		String path = "/v1/agent/check/pass/" + "check-" + serviceId;
 		try {
-			ConsulConnection consulConnection = getConnection(CHECK_PASS_CONNECTION_KEY);
+			ConsulConnection consulConnection = getConnection(CHECK_PASS_CONNECTION_KEY + Thread.currentThread().getId());
 			AtomicReference<ClientResponse> reference = consulConnection.send(Methods.PUT, path, token, null);
 			int statusCode = reference.get().getResponseCode();
 			if(statusCode >= UNUSUAL_STATUS_CODE){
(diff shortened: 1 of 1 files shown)


chenyan71:
It looks good. Thanks.

But if local thread is closed, the connection will be still in the cache map. Even the local thread ill be reused by JVM, but in special case the connection cache map could be group to big size:

	private ConcurrentHashMap<String, ConsulConnection> connectionPool = new ConcurrentHashMap<>();

I would like suggest to add a feature to clear the connection pool cache if it reach config size (for example, 1000), we clear the map and let it rebuild by new connection:

connectionPool.clear()

miklish:
> @miklish It looks good to me. I found a method needsToCreateConnection called from the send method to ensure that the connection is open before using it. Thanks.

@stevehu Yes, I saw that too. That is why I didn't add any code in getConnection to check if cached connections were still alive (as we had discussed yesterday).

chenyan71:
the default max work threads size is 200, maybe we are find with the cache size

miklish:
> It looks good. Thanks.
> 
> But if local thread is closed, the connection will be still in the cache map. Even the local thread ill be reused by JVM, but in special case the connection cache map could be group to big size:
> 
> ```
> private ConcurrentHashMap<String, ConsulConnection> connectionPool = new ConcurrentHashMap<>();
> ```
> 
> I would like suggest to add a feature to clear the connection pool cache if it reach config size (for example, 1000), we clear the map and let it rebuild by new connection:
> 
> connectionPool.clear()

If the HTTP connection inside the ConsulConnection object is not used for a long time, I believe either undertow or Consul will kill the connection. So only the ConsulConnection object will remain holding a closed connection.

If later (after the connection has been closed), and a thread requests that same ConsulConnection object, the ConsulConnection object first checks the status of the connection it holds, and will create a new one if needed.

miklish:
@chenyan71 @jsu216 Can you folks approve today? @chenyan71 You have a tight timeline... it appears you are OK with the changes. If you could approve asap it would be appreciated.

miklish:
@chenyan71 @jsu216 Nevermind, I see that this has been merged already :)

PR #738, opened 2020-07-03
fixes #723 Token replacement in config files does not allow special c…

### LINKED ISSUE #723: Token replacement in config files does not allow special characters
@chenyan71 and I have noticed that there is some non-obvious behaviour when config values are parsed in yaml files. Basically, it looks like the backslash '\\' and dollar sign '$' are not parsed as ordinary characters when placed in config files.

We've noticed this happen before in SQL Server queries that involve JSON-related functions as they use a dollar sign to reference the root object. We had two workarounds:

For a top-level entry, we escaped the dollar sign (\\$). E.g. `xyz.query: SELECT JSON_VALUE(abc, '\$.foo.bar') FROM def`

We also placed queries as elements in a list so that we didn't have to escape the dollar sign. E.g. 

```
xyz.queryMap:
   query1: SELECT JSON_VALUE(abc, '$.foo.bar') FROM def;
   query2: SELECT JSON_QUERY(abc, '$.foo.baz') FROM def;
```

We also had this recently happen for a `metrics.influxdbPass` value for which the encrypted password contained a dollar sign. From the stacktrace we saw, the error occurs at https://github.com/networknt/light-4j/blob/e56e0cf9c235c09bd1b7a6659c2cd31777f17f9f/config/src/main/java/com/networknt/config/ConfigInjection.java#L73.

stevehu:
…haracters

### DIFF
--- a/config/src/main/java/com/networknt/config/ConfigInjection.java
+++ b/config/src/main/java/com/networknt/config/ConfigInjection.java
@@ -70,7 +70,7 @@ public class ConfigInjection {
             if (!(value instanceof String)) {
                 return value;
             }
-            m.appendReplacement(sb, (String) value);
+            m.appendReplacement(sb, m.quoteReplacement((String)value));
         }
         return m.appendTail(sb).toString();
     }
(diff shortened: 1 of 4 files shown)


stevehu:
A new test case is added to get the value from values.yml and it works. Thanks a lot for the recommendation.

PR #741, opened 2020-07-04
fixes #740 check escape backslash before appy quoteReplacement config

### LINKED ISSUE #740: check escape backslash before appy quoteReplacement config
This is related to https://github.com/networknt/light-4j/issues/723 and the fixed is checked in already. However, some of our users have applied escape in the values.yml file and the blind fix will break their production if they decide to upgrade to the next release. After discussing the issue with our users, we found a workaround to avoid configuration changes for our users. We can double-check if the config value is escaped and only apply quoteReplacement if it is not escaped. 

Two more test cases have been added to demo the new change and I am sure that our users who have found a way to work around the original bug will happy :)

stevehu:


config/src/test/resources/config/password.yml
```diff
@@ -2,3 +2,5 @@
 # the config file not loaded properly. This is to reproduce the issue and get it fixed.
 password: ${password:abc$defg}
 value: ${password.escapeValue:abc$defg}
+sql1: ${password.sql1:SELECT}
+sql2: ${password.sql2:SELECT}
```
tsuilouis, line 6: Sorry, why there is just a SELECT here?

chenyan71: yes, it will use values.yml anyway

stevehu: exactly.

PR #755, opened 2020-08-25
fixes #753 handle text/plain and content type missing in BodyHandler

### LINKED ISSUE #753: handle text/plain and content type missing in BodyHandler
If the content type is text/plain, we consume the body stream and convert it into a string and attach it to the exchange. 
If content type is missing in the header, we just treat it as binary and put the inputStream into the exchange.

stevehu:


stevehu:
In the previous implementation, the BodyHandler will be passed if content-type is empty. In that case, the business handler must parse the body by itself which might be a burden for most users. After the change, we just put the input stream into the exchange and treat it as a binary stream. The business hander can just get the input stream from the exchange and consume it. It is the same process for an unknown content type.

ericbroda:
That makes sense, but I think we may want to document that pretty well.

stevehu:
You are right about the documentation. I have updated the BodyHandler doc to reflect the changes. Thanks.

https://github.com/networknt/light-doc/issues/194

Issue #840, opened 2020-11-24
AuditHandler cannot logged request, response, query parameters if response is sent by a framework middleware handler 


jiachen1120:
The current design of the audit handler is that users implement their own way to audit their requests and response in their own handler. 

However, it seems if the response is sent by framework middleware handler, people may be hard to auditing the request and response. For example, if openapi validation error happen, status will be sent back but this response won't be auditted.

Probably we can provide a option to users to audit these kind of request and response.

There are two situation:
1. Framework help users auditting every request and response, regardless of success or fail

- User can enable dump handler at this time. We can retrieve dumpped request and response from dumpped exchange attachment and add them into auditMap

2. Framework help users auditting failed request and response, and the response is sent by a framework middleware handler

- User set the auditOnError to true, then the response at this moment is the error status, which already exists in the auditInfo, we can directly add it to the auditMap

stevehu:
@jiachen1120 I would prefer the second option and still position the dump handler as a debugging tool for the dev environment as it is too heavy. @miklish What do you think?

jiachen1120:
@stevehu Yeah. Acctually the second situation already can fulfill our customer's requirements. Just wondering will it be useful if we also added dumpped request and response into auditMap in dev environment when people enable dumpped handler for debugging?

I personally think option2 should be enough. @miklish Looking forward to your opinion.

miklish:
I prefer option 2 as well @stevehu @jiachen1120

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [AuditHandler cannot logged request, response, query parameters if response is sent by a framework middleware handler light-4j 1.6.x](https://trello.com/c/ASKPTCjB/879-audithandler-cannot-logged-request-response-query-parameters-if-response-is-sent-by-a-framework-middleware-handler-light-4j-16x)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [AuditHandler cannot logged request, response, query parameters if response is sent by a framework middleware handler light-4j master](https://trello.com/c/bL4CvGcL/880-audithandler-cannot-logged-request-response-query-parameters-if-response-is-sent-by-a-framework-middleware-handler-light-4j-mast)

PR #849, opened 2020-12-03
- solved #840

jiachen1120:
issue: #840 
- audit request if bodyHandler enabled
- audit query parameters if request contains it
- audit response on error

### DIFF
--- a/audit/src/test/resources/audit.yml
+++ b/audit/src/test/resources/audit.yml
@@ -49,13 +49,19 @@ audit:
 - serviceId
 
 # Request Body, this is optional and must be set by the service in its implementation
-- request
+- requestBody
 
 # Response payload, this is optional and must be set by the service in its implementation
-- response
+- responseBody
 
 # Request query parameters
 - queryParameters
+
+# Request path parameters
+- pathParameters
+
+# Request cookies
+- requestCookies
 
 # Response error status payload
 - Status
--- a/audit/src/test/resources/audit.yml
+++ b/audit/src/test/resources/audit.yml
@@ -1,0 +1,61 @@
+# AuditHandler will pick some important fields from headers and tokens and logs into a audit appender.
+---
+# Enable Audit
+enabled: true
+
+# Enable mask
+mask: false
+
+# Output response status code
+statusCode: true
+
+# Output response time
+responseTime: true
+
+# audit only on error responses; by default log for any return status code
+# if true, audit at error level if status code >= 400
+# log level is controlled by logLevel
+auditOnError: false
+
+# log level; by default set to info
+logLevelIsError: false
+
+
+# Output header elements. You can add more if you want.
+headers:
+
+# Correlation Id
+- X-Correlation-Id
+
+# Traceability Id
+- X-Traceability-Id
+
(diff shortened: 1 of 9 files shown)


stevehu (approved):
I have reviewed the PR and it looks good to me. @miklish Could you please review and approve it? Thanks.

audit/src/test/resources/audit.yml
```diff
@@ -0,0 +1,61 @@
...
+- scope_client_id
+
+# Request endpoint uri@method.
+- endpoint
+
+# Service ID assigned to the service, this is optional and must be set by the service in its implementation
+- serviceId
+
+# Request Body, this is optional and must be set by the service in its implementation
+- request
```
jaswalkiranavtar, line 52: can we rename this to requestBody and responseBody? Just saying request  sounds like it will audit the whole request including query params, path params, cookies etc

audit/src/test/resources/audit.yml
```diff
@@ -0,0 +1,61 @@
...
+
+# Response payload, this is optional and must be set by the service in its implementation
+- response
```
jaswalkiranavtar, line 55: same comment as comment for request

audit/src/test/resources/audit.yml
jiachen1120: Thanks for pointing out, I also think renaming would be better. But seems these names have been used in previous versions. If we change them now, it may cause backward compatibility. People who previously used these names to audit requests may not be able to continue auditing them if they don't change their configuration. @stevehu Do you think we can change it?

audit/src/test/resources/audit.yml
```diff
@@ -0,0 +1,61 @@
...
+
+# Request query parameters
+- queryParameters
```
jaswalkiranavtar, line 58: you added queryParams, what if somebody wants to audit path params and cookies?

audit/src/test/resources/audit.yml
stevehu: I think it make sense to rename it to clearly define what is the audited object. Regarding the backward compatibility in the config, we can mention that in the next release notes. It is not a coding issue but only the configuration file and it is easier for users to adjust. Thank you for pointing it out.

jiachen1120: I see. Will rename them.

audit/src/test/resources/audit.yml
jiachen1120: Thanks for pointing out. 

Since my client now only needs audit query parameters, I was limited to audit query parameters before. I think we can try to cover more audit point.

stevehu: I have realized that serviceId is in the audit list but it is not implemented. It is only mentioned in the comment section at the beginning of the handler. When the audit log aggregated to a central location, the serviceId should be the key info. @jiachen1120 Could you please add this to this PR? It should be retrieved from the server.yml file. Let me know if you have questions or concerns.

jiachen1120: Sure. Will do

stevehu (approved):

PR #860, opened 2020-12-11
issue-858

BalloonWen:
- added timestampFormat in audit.yml, if timestampFormat is specified, then will use the format, 
if the format of the timestamp is invalid, it will use a long value.
- if the format is incorrect, it will show an error when the server starts up

### DIFF
--- a/audit/src/main/resources/config/audit.yml
+++ b/audit/src/main/resources/config/audit.yml
@@ -20,7 +20,7 @@ auditOnError: false
 # log level; by default set to info
 logLevelIsError: false
 
-# the format for outputting the timestamp, if not specify, will use long value
+# the format for outputting the timestamp, if the format is not specified or invalid, will use a long value.
 #timestampFormat: yyyy-MM-dd'T'HH:mm:ss.SSSZ
 
 # Output header elements. You can add more if you want.
--- a/audit/src/main/resources/config/audit.yml
+++ b/audit/src/main/resources/config/audit.yml
@@ -20,6 +20,8 @@ auditOnError: false
 # log level; by default set to info
 logLevelIsError: false
 
+# the format for outputting the timestamp, if not specify, will use long value
+#timestampFormat: yyyy-MM-dd'T'HH:mm:ss.SSSZ
 
 # Output header elements. You can add more if you want.
 headers:
--- a/audit/src/test/java/com/networknt/audit/AuditHandlerTest.java
+++ b/audit/src/test/java/com/networknt/audit/AuditHandlerTest.java
@@ -62,8 +62,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
 
 /**
  * Created by steve on 01/09/16.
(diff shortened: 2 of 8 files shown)


jaswalkiranavtar:
What will happen if user accendtly provide invalid format? Can we stop the app from starting or fallback to some default format and signal the user that you entered invalid format that is why he are seeing fallback format?

BalloonWen:
> What will happen if user accendtly provide invalid format? Can we stop the app from starting or fallback to some default format and signal the user that you entered invalid format that is why he are seeing fallback format?

I can add a log if there's an invalid format, because in production, the format should be a tested one.
If we fallback to  a format, it will add confusion to developer why the format doesn't change or so.
To stop the server, we could do that, but then, the server will depends on audit handler, which I don't think it's a good approach based on current structure.

What do you think?

stevehu:
We cannot stop the server but should output an error status code designate for this error. It should be very easy to detect using the development and testing cycle and get it fixed.

audit/src/test/java/com/networknt/audit/AuditHandlerTest.java
```diff
@@ -358,6 +359,38 @@ public void testAuditWith401ServiceId() throws Exception {
...
+        long time = 1607639411945L;
+        Instant instant = Instant.ofEpochMilli(time);
+        PowerMockito.mockStatic(Instant.class);
+        PowerMockito.when(Instant.now()).thenReturn(instant);
+
+        runTest("/error", "post", null, 401);
+        verifyAuditInfo("timestamp", "2020-12-10T17:30:11.945-0500");
+    }
+
+/*    @Test used for testing when doesn't specify timestampFormat
```
jaswalkiranavtar, line 384: shall we delete this, if this test is not being used?

audit/src/test/java/com/networknt/audit/AuditHandlerTest.java
```diff
@@ -358,6 +359,38 @@ public void testAuditWith401ServiceId() throws Exception {
         verifyAuditInfo("serviceId", "com.networknt.petstore-1.0.0");
     }
 
+    @Test
+    public void testAuditWith200TimestampFormatted() throws Exception {
```
jaswalkiranavtar, line 363: Is it also possible to test the situation when adds the invalid format, with current solution, I would expect the app failing to start.

BalloonWen: > Is it also possible to test the situation when adds the invalid format, with current solution, I would expect the app failing to start.

I will add a test for that, but the server won't stop because of the format error.

audit/src/main/resources/config/audit.yml
```diff
@@ -20,6 +20,8 @@ auditOnError: false
 # log level; by default set to info
 logLevelIsError: false
 
+# the format for outputting the timestamp, if not specify, will use long value
```
jaswalkiranavtar, line 23: lets also document here what would happen if the format that user entered is invalid.

audit/src/test/resources/audit.yml
```diff
@@ -20,6 +20,9 @@ auditOnError: false
 # log level; by default set to info
 logLevelIsError: false
 
+# the format for outputting the timestamp
```
jaswalkiranavtar, line 23: lets also document here what would happen if the format that user entered is invalid.

BalloonWen: feat/858-audit-format

jaswalkiranavtar (approved):
[LGTM](https://lgtm.lol/i/433)

stevehu (approved):

PR #863, opened 2020-12-11
Fix/861 audit service

BalloonWen:
fixed issue when getJsonMapConfig


jiachen1120:
@BalloonWen Thanks for the quick fix!! Do you think it would be better to load the configuration when class initialize to prevent load server config for every requests? 

for example load when construct audithandler?

```java
public AuditHandler() {
        if (logger.isInfoEnabled()) logger.info("AuditHandler is loaded.");
        auditConfig = AuditConfig.load();
        serverConfig = // load no cache config here?
    }
```

stevehu:
@jiachen1120 I think that is a good idea as we are not using cache anymore. In the Http2Client, I am loading the server.yml in a static block.

BalloonWen:
@jiachen1120 good idea, @stevehu will do in this PR

PR #865, opened 2020-12-14
issue-864

BalloonWen:
- now logging the status ans stack trace separately
- now will only log the stack trace when TRACE is enabled.

### DIFF
--- a/handler/src/main/java/com/networknt/handler/LightHttpHandler.java
+++ b/handler/src/main/java/com/networknt/handler/LightHttpHandler.java
@@ -28,6 +28,7 @@ import io.undertow.util.Headers;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


handler/src/main/java/com/networknt/handler/LightHttpHandler.java
```diff
@@ -43,24 +44,24 @@
...
- 	boolean auditOnError = config != null ? config.getAuditOnError() : false;
- 	boolean auditStackTrace = config != null ? config.getAuditStackTrace() : false;
-    
     /**
      * This method is used to construct a standard error status in JSON format from an error code.
      *
      * @param exchange HttpServerExchange
-     * @param code error code
-     * @param args arguments for error description
+     * @param code     error code
```
jaswalkiranavtar, line 59: no need for tab here as per javadoc standard

handler/src/main/java/com/networknt/handler/LightHttpHandler.java
```diff
@@ -43,24 +44,24 @@
...
+     * @param args     arguments for error description
```
jaswalkiranavtar, line 60: no need for tab here as per javadoc standard

handler/src/main/java/com/networknt/handler/LightHttpHandler.java
```diff
@@ -72,27 +73,35 @@ default void setExchangeStatus(HttpServerExchange exchange, String code, final O
      * want to bubble up to the caller and eventually to the original caller.
      *
      * @param exchange HttpServerExchange
-     * @param status error status
+     * @param status   error status
```
jaswalkiranavtar, line 76: no need for tab here as per javadoc standard

handler/src/main/java/com/networknt/handler/LightHttpHandler.java
BalloonWen: This is for align with the HttpServerExchange
Based on the example on Oracle doc https://www.oracle.com/ca-en/technical-resources/articles/java/javadoc-tool.html#format
You can see the "url" "name " in this example is also aligned.
I guess either aligned or not aligned is fine, what do you think?

jaswalkiranavtar: Ok I am fine if according to official documentation, it is acceptable. But i have not seen it much in practice.
Looks good to me.
Apprving

jaswalkiranavtar (approved):

stevehu (approved):

PR #879, opened 2020-12-22
Enhancing to use different env variables in different junit test classes

jaswalkiranavtar:
System.getenv() and System.getenv(String) uses ProcessEnvironment.getenv() and ProcessEnvironment.getenv(String) respectively. If you take a look at ProcessEnvironment.getenv() and ProcessEnvironment.getenv(String), you will find that they return different maps and because of that we may get different values when invoking both these methids. Hence changing line 99 in ConfigInjection to make it same as 110.

With this fix, I am able to set different environment variables in different junit test classes.

image

### DIFF
--- a/config/src/main/java/com/networknt/config/ConfigInjection.java
+++ b/config/src/main/java/com/networknt/config/ConfigInjection.java
@@ -96,7 +96,7 @@ public class ConfigInjection {
             // Flag to validate whether the environment or values.yml contains the corresponding field
             Boolean containsField = false;
             // Use key of injectionPattern to get value from both environment variables and "values.yaml"
-            Object envValue = typeCast(System.getenv(injectionPattern.getKey()));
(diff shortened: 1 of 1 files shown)


stevehu:
@jaswalkiranavtar @BalloonWen Based on our offline conversation, the behaviour is the same for the platform as theUnmodifiableEnvironment is just a read-only snapshot of theCaseInsensitiveEnvironment. When we are trying to manipulate the environment variable in Unit Tests, we just need to remember to manipulate the right map in order for the config module to pick it up. Based on the explanation from Bolun, getting env with a string key has the advantage to differentiate between null value vs key doesn't exist. So we will keep the current implementation and change the test case to manipulate the theCaseInsensitveEnvironment for injections. We need to document this clearly in our config module for the environment injection in JUnit test though.

PR #889, opened 2020-12-31
Feat/#884 add metadata to Status

BalloonWen:
added metadata to status

### DIFF
--- a/handler/src/main/java/com/networknt/handler/LightHttpHandler.java
+++ b/handler/src/main/java/com/networknt/handler/LightHttpHandler.java
@@ -74,6 +74,24 @@ public interface LightHttpHandler extends HttpHandler {
     }
 
     /**
+     * This method is used to construct a standard error status with metadata in JSON format from an error code.
+     *
+     * @param exchange HttpServerExchange
+     * @param code     error code
+     * @param metadata additional metadata info
+     * @param args     arguments for error description
+     */
+    default void setExchangeStatus(HttpServerExchange exchange, String code, Map<String, Object> metadata, final Object... args) {
+        Status status = new Status(code, args);
+        if (status.getStatusCode() == 0) {
+            // There is no entry in status.yml for this particular error code.
+            status = new Status(ERROR_NOT_DEFINED, code);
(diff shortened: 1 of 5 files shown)


status/src/main/java/com/networknt/status/Status.java
```diff
@@ -258,7 +258,7 @@ public String toString() {
             if (showDescription) {
                 sb.append("\",\"description\":\"" + getDescription());
             }
-            if (showMetadata) {
+            if (showMetadata && getMetadata() != null) {
                 try {
                     sb.append("\",\"metadata\":\"" + Config.getInstance().getMapper().writeValueAsString(getMetadata()));
```
stevehu, line 267: When we convert the metadata map to the string, is the final result still can be parsed as a JSON?

handler/src/main/java/com/networknt/handler/LightHttpHandler.java
```diff
@@ -73,6 +73,24 @@ default void setExchangeStatus(HttpServerExchange exchange, String code, final O
...
 
+    /**
+     * This method is used to construct a standard error status with metadata in JSON format from an error code.
+     *
+     * @param exchange HttpServerExchange
+     * @param code     error code
+     * @param metadata additional metadata info
+     * @param args     arguments for error description
+     */
+    default void setExchangeStatus(HttpServerExchange exchange, String code, Map<String, Object> metadata, final Object... args) {
```
stevehu, line 84: Should we call the previous method to reduce the code duplication?

status/src/main/java/com/networknt/status/Status.java
BalloonWen: now the metadata is appending a JSON instead of a string, and it can be parsed back.

stevehu:
@BalloonWen Could you please update the document for the status with the latest changes? Thanks.

BalloonWen:
> @BalloonWen Could you please update the document for the status with the latest changes? Thanks.

Please take a look and advise https://github.com/networknt/light-doc/pull/234

PR #899, opened 2021-01-26
Feat/#898 oauth key

### LINKED ISSUE #898: add Status error when cannot resolve Oauth key
Need add a error status when there's a exception to get the certificate from key, so that the user will easier to track this error by error code instead of a generic Runtime Exception

BalloonWen:
resolve #898 
simply log a status when get Oauth key fail.
this will be better for error tracking purpose

### DIFF
--- a/status/src/main/resources/config/status.yml
+++ b/status/src/main/resources/config/status.yml
@@ -447,6 +447,11 @@ ERR10064:
   code: ERR10064
   message: EMPTY_LAMBDA_RESPONSE
   description: Empty lambda function response for %s.
+ERR10065:
+  statusCode: 400
+  code: ERR10065
+  message: OAUTH_KEY_ERROR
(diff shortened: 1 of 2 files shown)


stevehu (approved):

status/src/main/resources/config/status.yml
```diff
@@ -447,6 +447,11 @@ ERR10064:
   code: ERR10064
   message: EMPTY_LAMBDA_RESPONSE
   description: Empty lambda function response for %s.
+ERR10065:
+  statusCode: 400
+  code: ERR10065
+  message: OAUTH_KEY_ERROR
+  description: Error happened when getting Oauth key.
```
jaswalkiranavtar, line 454: please rename to and also do not end with ".":
"Unable to download jwk set"

PR #913, opened 2021-02-10
allow key injection in configuration

### LINKED ISSUE #912: Config should also inject env variable / values.yml to config key
/health/liveness/${service_id}: api0061

service_id could also be injected

BalloonWen:
resolved #912
- allow key injection in configuration

### DIFF
--- a/config/src/test/resources/config/testKeyInject.yml
+++ b/config/src/test/resources/config/testKeyInject.yml
@@ -1,0 +1,11 @@
+#key1 test inject key with default value, test inject key of Map type
+${key1}:
+  #key2 test inject key with content other than inject pattern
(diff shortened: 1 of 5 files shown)


config/src/test/resources/config/testKeyInject.yml
```diff
@@ -0,0 +1,11 @@
+#key1 test inject key with default value, test inject key of Map type
```
jaswalkiranavtar, line 1: please ass this test case, as this is our scenario:

key1${key1}: val1

jaswalkiranavtar (approved):

stevehu (approved):

Issue #897, opened 2021-01-25
resolve oauth key when start up 


BalloonWen:
resolve key when start

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [resolve oauth key when start up light-4j master](https://trello.com/c/Hv6I6vM5/947-resolve-oauth-key-when-start-up-light-4j-master)

PR #918, opened 2021-02-17
issue #897 key resolving at the start up

BalloonWen:
- now when use JWK resolver, it will get the key during the start up and cache
- when handling request, if it cannot find the corresponding kid (probably due to key rotation), it will try to get from the OAuth server again


security/src/main/java/com/networknt/security/JwtVerifier.java
```diff
@@ -322,36 +329,34 @@ private VerificationKeyResolver getKeyResolver(String kid, boolean isToken) {
...
                 }
+                logger.debug("Got Json web key set for kid: {} from local cache", kid);
+                verificationKeyResolver = new JwksVerificationKeyResolver(jwkList);
                 break;
         }
         return verificationKeyResolver;
     }
 
     /**
      * Retrieve JWK set from oauth server with the given kid
```
jaswalkiranavtar, line 348: change the description, it is not for a given kid anymore

stevehu (approved):

PR #928, opened 2021-04-13
Added code to set MDC data

mayurikasaxena:


server/src/main/java/com/networknt/server/Server.java
```diff
@@ -116,6 +124,10 @@ public static void init() {
 
             // this will make sure that all log statement will have serviceId
             MDC.put(SID, config.getServiceId());
+
```
BalloonWen, line 127: correlation id may be not needed when the server starts, since correlation id is specific to the requests.

server/pom.xml
```diff
@@ -109,6 +109,10 @@
             <artifactId>testcontainers</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
BalloonWen, line 112: if the correlation id is not needed when server starts, then the dependency will not be needed.

correlation/src/main/java/com/networknt/correlation/CorrelationHandler.java
```diff
@@ -59,21 +67,42 @@ public CorrelationHandler() {
...
         if(cId == null) {
         	// if not set, check the autgen flag and generate if set to true
         	if(config.isAutogenCorrelationID()) {
 	            // generate a UUID and put it into the request header
 	            cId = Util.getUUID();
 	            exchange.getRequestHeaders().put(HttpStringConstants.CORRELATION_ID, cId);
 	            String tId = exchange.getRequestHeaders().getFirst(HttpStringConstants.TRACEABILITY_ID);
+	            
+	            MDC.put(CID, cId);
+	            if(!"".equalsIgnoreCase(serviceId)) {
```
BalloonWen, line 86: This part I'm not sure, @stevehu  should we put service id in the correlation handler? it seems mix the feature, but will also be too much to create a module just for putting the service id?

stevehu (commented):
I don't think we should put the serviceId into the MDC as it is a constant for a particular service. The MDC is based on the thread and we should only add context info per request as they are changing per thread. For serviceId, we can update the logback.xml to add a property there to output in the logging statement. In your case when Splunk is used, the best way is to associate the serviceId to the Splunk data loader to inject the serviceId to the Splunk during the data loading. What do you think?

mayurikasaxena:
Change/Ticket was not required.

Issue #936, opened 2021-04-26
When request payload is an invalid json, AuditHandler fails to log the request


jsu216:
AuditHandler relies on the BodyHandler (or maybe other handler) to record request payload (as attachment) first. When BodyHandle failed to parse the request, nothing can be recorded in audit log.
For an example, the following request body will get a valid response but no audit log will be created.
{
    "key1": 
        "key2": 1234567890
    }
}
Also, the "auditRequest" is outside of the exchangeCompleteListener. So whenever an exception thrown by a handler defined before the AuditHandler, the request cannot be logged in audit. BodyHandler is just an example.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [When request payload is an invalid json, AuditHandler fails to log the request](https://trello.com/c/Fr8TJqJ2/1020-when-request-payload-is-an-invalid-json-audithandler-fails-to-log-the-request)

stevehu:
The master branch uses the request and response interceptor to capture the request body and response body. The 1.6.x is merged with the fix. Close it.

Issue #935, opened 2021-04-26
When mask is turned on, AuditHandler fails to log the request


jsu216:
When the request body is a valid json object, the BodyHandler converts it to a java object (Map or List).
Since masking cannot be done on the object directly, the AuditHandler converts it to a String first. But it calls the toString() method of the object instead of the reserialize it. The result of the toString() method is not a valid Json format (e.g. missing double quote for key name). So this stops the masking and no audit log recorded.

There is an outstanding issue related to this [#13](https://github.com/networknt/light-4j/issues/13)

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [When mask is turned on, AuditHandler fails to log the request  light-4j 1.6.x](https://trello.com/c/P3miFowF/1021-when-mask-is-turned-on-audithandler-fails-to-log-the-request-light-4j-16x)

PR #937, opened 2021-04-28
Fix audit issue #935 and #936

jsu216:
Fix the two issues #935 and #936

### DIFF
--- a/audit/src/main/java/com/networknt/audit/AuditHandler.java
+++ b/audit/src/main/java/com/networknt/audit/AuditHandler.java
@@ -103,26 +103,26 @@ public class AuditHandler implements MiddlewareHandler {
 
     @Override
     public void handleRequest(final HttpServerExchange exchange) throws Exception {
-        Map<String, Object> auditInfo = exchange.getAttachment(AttachmentConstants.AUDIT_INFO);
         Map<String, Object> auditMap = new LinkedHashMap<>();
         final long start = System.currentTimeMillis();
         auditMap.put(TIMESTAMP, System.currentTimeMillis());
 
-        // dump audit info fields according to config
-        boolean needAuditData = auditInfo != null && auditConfig.hasAuditList();
-        if (needAuditData) {
-            auditFields(auditInfo, auditMap);
-        }
-
-        // dump serviceId from server.yml
-        if (auditConfig.hasAuditList() && auditConfig.getAuditList().contains(SERVICEID_KEY)) {
-            auditServiceId(auditMap);
-        }
-
         if (auditConfig.isStatusCode() || auditConfig.isResponseTime()) {
             exchange.addExchangeCompleteListener((exchange1, nextListener) -> {
+                Map<String, Object> auditInfo = exchange.getAttachment(AttachmentConstants.AUDIT_INFO);
+                // dump audit info fields according to config
+                boolean needAuditData = auditInfo != null && auditConfig.hasAuditList();
+                if (needAuditData) {
+                    auditFields(auditInfo, auditMap);
+                }
+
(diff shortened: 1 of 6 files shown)


miklish:
FYI: We are waiting for our client to review and test these changes before submit any approvals

chenyan71:
I add some concerns below for this PR, not sure if it is valid:

1. why auditRequest need statusCode and responseTime to be true in the config? (it moved into if block):
          if (auditConfig.isStatusCode() || auditConfig.isResponseTime()) {
            exchange.addExchangeCompleteListener((exchange1, nextListener) -> {
                auditRequest(**exchange**, auditMap, auditConfig);
    Should it be:   auditRequest(**exchange1**, auditMap, auditConfig); ?

2. for the requestBody method change, it make sense to log the  json （application/json） to audit log as issue m#935 mentioned.
    But if the json convert failed by Objectmapper with exception：

                try {
                    requestBodyString = Config.getInstance().getMapper().writeValueAsString(requestBody);
                } catch (JsonProcessingException e) {
                    logger.error("Failed to audit log request body", e);
                }

   The requestBodyString  will be still  null and handler will skip next if statement (if (requestBodyString != null)) . So the audit log will still not have invalid json string for user to audit(I am not sure if it is PR requestor wanted). So I suggest we integrate two if statements together. When exception happen, add the raw string to audit log.

3. for the binary format requestbody (InputStream for pdf, image, ...), do we need add it to audit log? It may mess the audit log.
  content-type not in ("application/json", "text/plain", "multipart/form-data", "application/x-www-form-urlencoded")

jsu216:
Thanks @chenyan71 for your comment. My replies are below and I will make some changes to the PR accordingly.

> I add some concerns below for this PR, not sure if it is valid:
> 
> 1. why auditRequest need statusCode and responseTime to be true in the config? (it moved into if block):
>    if (auditConfig.isStatusCode() || auditConfig.isResponseTime()) {
>    exchange.addExchangeCompleteListener((exchange1, nextListener) -> {
>    auditRequest(**exchange**, auditMap, auditConfig);
>    //Should it be:   auditRequest(**exchange1**, auditMap, auditConfig); ?

This PR does not change condition when the "addExchangeCompleteListener" is set, but only moves the "auditRequest" and others into the listener. I agree that this changes the condition that these audit functions being called. I will move the status and responsetime  check into the listener as well.

> 2. for the requestBody method change, it make sense to log the  json （application/json） to audit log as issue m#935 mentioned.
>    But if the json convert failed by Objectmapper with exception：
>    ```
>             try {
>                 requestBodyString = Config.getInstance().getMapper().writeValueAsString(requestBody);
>             } catch (JsonProcessingException e) {
>                 logger.error("Failed to audit log request body", e);
>             }
>    ```
>    
>    
>    The requestBodyString  will be still  null and handler will skip next if statement (if (requestBodyString != null)) . So the audit log will still not have invalid json string for user to audit(I am not sure if it is PR requestor wanted). So I suggest we integrate two if statements together. When exception happen, add the raw string to audit log.

This only happens when the REQUEST_BODY attachment is populated as an object but cannot be serialized as a Json String.
I will make the change to use Object.toString() as the requestBodyString to avoid missing it in audit log.

> 3. for the binary format requestbody (InputStream for pdf, image, ...), do we need add it to audit log? It may mess the audit log.
>    content-type not in ("application/json", "text/plain", "multipart/form-data", "application/x-www-form-urlencoded")

This is not in the scope of this PR. We need another issue or feature to address this.

chenyan71:
@jsu216  Thanks for the update. Please ignore the item 3.
I will create a separate issue for it and ask @stevehu  to verify.

zabooma (approved):

stevehu (approved):

PR #969, opened 2021-06-25
fixes #965

### LINKED ISSUE #965: add ingress-proxy and egress-router modules
We have the light-router for legacy client/consumer to invoke light-4j APIs and light-proxy to bring legacy services to the light-4j ecosystem. Now we are working on the light-mesh/http-sidecar to combine both light-router and light-proxy cross-cutting concerns running in the same k8s pod to address cross-cutting concerns for backend API build with any frameworks including light-4j. 

To promote code reuse and reduce the duplications, we need to move the cross-cutting concerns from the light-router and light-proxy to the light-4j in order to share them with light-mesh/http-sidecar. 

For some light-router and light-proxy specific middleware handlers that are not going to be used by the light-mesh, we can keep them in these modules.

chenyan71:
add two new modules: one for router handlers and another one for proxy handlers.

The light-router / light-proxy /light-mesh will use dependency on these modules instead of have the handlers in the project

### DIFF
--- a/egress-router/README.md
+++ b/egress-router/README.md
@@ -1,0 +1,20 @@
+## egress-router module
+
+The module provides router (egress traffic) related handlers and util classes.
+
+ 
+
+When the system need call light-api from legacy system or API developed different technologies, for example, node.js, .net, the client side need process service discovery, JWT token management, load TLS certs...;
+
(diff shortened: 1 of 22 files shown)


stevehu (approved):

stevehu:
I have done a snapshot release for this update. Thanks.

PR #997, opened 2021-07-11
fix light-rest-4j issue #183

chenyan71:
add interface for config load in case user want to skip the cache and load config from config file directly


chenyan71:
The method already existing, close the PR

PR #1005, opened 2021-07-18
 build error fix, change egress_router value scope

chenyan71:
- A test case error fix, 
change the "next handler" for egress to protected to allow extend class use it


stevehu:
[build error fix, change egress_router value scope light-4j master](https://trello.com/c/zBVt01ay/1130-build-error-fix-change-egressrouter-value-scope-light-4j-master)

PR #1020, opened 2021-08-07
 fixes #1019

### LINKED ISSUE #1019: Router Token handler enhancement
Currently, the Router Token handler will always add new generated token to AUTHORIZATION header

        exchange.getRequestHeaders().put(Headers.AUTHORIZATION, "Bearer " + cachedJwt.getJwt());

But if the request has subject token, it should be keep and the new token should be added  as Scope_token

GavinChenYan:


GavinChenYan:
@stevehu  Please take a look if we should apply same logic in SAMLTokenHandler

stevehu (approved):

stevehu:
I think we should do the same for the saml token handler.

PR #1022, opened 2021-08-11
 add status change and new method for Result object

GavinChenYan:
Add status check, include customer define accept status
Add new method to return Result Object


stevehu:
[add status change and new method for Result object light-4j master](https://trello.com/c/6Y7JByjt/1158-add-status-change-and-new-method-for-result-object-light-4j-master)

PR #1025, opened 2021-08-31
add caller id header to audit map which will be used metrics

GavinChenYan:


stevehu (approved):

stevehu:
[add caller id header to audit map which will be used metrics light-4j master](https://trello.com/c/elCcXd4d/1189-add-caller-id-header-to-audit-map-which-will-be-used-metrics-light-4j-master)

PR #1028, opened 2021-09-03
Inefficient Usages of Java Collections

FastAtlas:
Hi,
We find that the List allocated at line 85 in the function `isExclusionConfigFile` is only used to check whether an element is stored in the container. The method `contains` of a list runs in linear time complexity. We notice that the `exclusionFileList` is maintained in the HashMap `exclusionMap`, which maps a string object to an object.

To achieve the same functionality, maybe a `HashSet` can provide a more efficient implementation. We can just construct a HashSet and maintain it in the HashMap `exclusionMap`. When we invoke the function `isExclusionConfigFile`, we only need to take amortized O(1) time to maintain the map and invoke the `contains` method of `HashSet`.

We discovered this inefficient usage of containers by our tool Ditto. The patch is submitted. Could you please check and accept it? We have tested the patch on our PC. The patched program works well. Of course, it would be better to change `EXCLUSION_CONFIG_FILE_LIST` to `EXCLUSION_CONFIG_FILE_Set`. Anyway, the name does not affect the performance anymore.

Bests

Ditto

### DIFF
--- a/config/src/main/java/com/networknt/config/ConfigInjection.java
+++ b/config/src/main/java/com/networknt/config/ConfigInjection.java
@@ -82,10 +82,16 @@ public class ConfigInjection {
     // Return the list of exclusion files list which includes the names of config files that shouldn't be injected
     // Double check values and exclusions to ensure no dead loop
     public static boolean isExclusionConfigFile(String configName) {
-        List<Object> exclusionConfigFileList = (exclusionMap == null) ? new ArrayList<>() : (List<Object>) exclusionMap.get(EXCLUSION_CONFIG_FILE_LIST);
+        HashSet<Object> exclusionConfigFileSet;
+        if (exclusionMap == null || !exclusionMap.containsKey(EXCLUSION_CONFIG_FILE_LIST)) {
+            exclusionConfigFileSet = new HashSet<>();
+        } else {
+            exclusionMap.put(EXCLUSION_CONFIG_FILE_LIST, new HashSet(exclusionMap.get(EXCLUSION_CONFIG_FILE_LIST)));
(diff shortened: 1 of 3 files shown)


stevehu:
For any use case that container more data elements in the list, it would make sense to convert the list to set to speed up the contain calculation, however, in the case of exclusive list, we only have about 4 elements in the list. It takes more time to convert it to set than walking through the 4 elements. Also, it makes the logic harder for other developers to understand. I think we shouldn't take the performance optimization as the number of elements is too small. What do you think?

DittoTool:
Steve, I generally agree with you. I refined the patch and removed unnecessary construction of HashSet objects. The set is only constructed one time, so the concern you mentioned above would not exist.

We detected this inefficient usage by a static analyzer and do not analyze the number of elements maintained in the list in the actual use. When the size of the list is small, the transformation will not bring much difference.

Anyway, we discovered this usage and recommended avoiding it in the design. For example, the type of `exclusionMap` can be restricted to `HashMap<String, HashSet>` rather than `HashMap<String, Collection>` if possible. Programmers tend to declare the type of containers in a rough way. However, it would be better for performance improvement when we consider how the container objects are utilized in the program. The same functionality can be achieved by other container types and interfaces. Of course, the patch might degrade the cleanness of the code.

DittoTool:
Similarly, we also find other two cases of inefficient usages of List, which can be replaced by HashSet objects. Meanwhile, two ArrayList can be replaced by LinkedList. The two ArrayList objects are inserted before iterations, while multiple memory reallocation might occur when the size of the list exceeds its capacity. We recommend replacing it with a LinkedList.

PR #1032, opened 2021-09-16
add default header when client request missed the setting

GavinChenYan:
for the POST request, if the request header doesn't have following values:
-CONTENT_TYPE
-TRANSFER_ENCODING

set default


stevehu (approved):

stevehu:
[add default header when client request missed the setting light-4j master](https://trello.com/c/XNFwf6ro/1223-add-default-header-when-client-request-missed-the-setting-light-4j-master)

PR #1056, opened 2021-11-03
Issue1055 - ClassCastException in maskList method

### LINKED ISSUE #1055: ClassCastException in maskList method fix in light-4j version 1.6.X
There was an issue #942 raised related to handling one element in an array for masking.  PR #945 was created for the same and the fix was included in the subsequent light-4j version `2.0.27`.

Our client, who is using the light-4j `1.6.X` version, has raised exactly the same issue. The fix done for #942 needs to be applied and pushed into the upcoming light-4j `1.6.X` version.

jaydeepparekh1311:
Fixes #1055 to handle only one element in an array for masking

### DIFF
--- a/mask/src/main/java/com/networknt/mask/Mask.java
+++ b/mask/src/main/java/com/networknt/mask/Mask.java
@@ -223,11 +223,18 @@ public class Mask {
          */
         if(pathList != null && pathList.size() == 1) {
             String path = pathList.get(0);
-            List values = ctx.read(path);
+            // the value might be a string or a list
+            Object readValues = ctx.read(path);
+            List values = new ArrayList();
+            if(readValues instanceof String){
+                values.add(readValues);
+            } else {
+                values = ctx.read(path);
+            }
             List maskedValue = new ArrayList();
             //mask each value in the list of the same path
             values.forEach(o -> maskedValue.add(replaceWithMask(o.toString(), MASK_REPLACEMENT_CHAR.charAt(0), expression)));
-            ctx.set(path, maskedValue);
+            ctx.set(path, maskedValue.size() == 1 ? maskedValue.get(0) : maskedValue);
         } else {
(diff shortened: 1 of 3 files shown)


stevehu (approved):

miklish:
I would suggest naming the new unit test methods using names that reflect the problem, not ones that reference the abstract issue number. (See `MaskTest.java` (`testMaskIssue1055(..)`) and `mask.yml` (`testIssue1055:`). Instead I'd suggest something like `testMaskSingleElementList(..)` and `testSingleElementList:`.

Also, this is nitpicking, but (1) the creation of the ArrayList is not always needed, (2) this code also calls `ctx.read(..)` twice unnecessarily (assuming `ctx.read(path)` only ever returns a `String` or a `List`). So instead of:

```java
            // the value might be a string or a list
            Object readValues = ctx.read(path);
            List values = new ArrayList();
            if(readValues instanceof String){
                values.add(readValues);
            } else {
                values = ctx.read(path);
            }
```

Instead it could be:

```java
            // readValues might be a String or a list
            Object readValues = ctx.read(path);
            List values;
            if(readValues instanceof String)
                (values = new ArrayList()).add(readValues);
            else
                values = (List) readValues;
```

But this is just nitpicking as mentioned :)

miklish:
@stevehu @jaydeepparekh1311 Comments?

stevehu:
@miklish @jaydeepparekh1311 I don't have a preference for the test case implementation. If this is the main code, I would go with @miklish's implementation. I will let you guys work out the final version and start the 1.6.x release. Thanks.

jaydeepparekh1311:
Thank you @miklish and @stevehu for your review comments. Agree with the same. PR #1063  is created with the updated code changes. Also, I have added you'll as reviewers for the PR #1063. 

Can you please review and approve the same to be merged into branch 1.6.x, incase everything looks good?

PR #1078, opened 2021-12-02
Feature/dependency update

KalevGonvick:
**Updated Apache Commons Codec from 1.11 -> 1.13**
- Fixes Apache Commons Invalid Base32 String Tunneling Weakness.

**Updated Logback from 1.2.3 -> 1.2.7**
- Fixes *some* Logback Config File Field Server-side Request Forgery (SSRF) weaknesses.

**Additional Notes**
Logback 1.2.7 still has more issues with request forgery, so we should update to 1.3.0 when it gets a full release (it is still 1.3.0-alpha).

The Apache Commons Compress dependency has a known security risk with an exception DDOS exploit. The newest release does not fix this issue, so we should keep an eye out for an update.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,6 @@
         <version.jose4j>0.6.3</version.jose4j>
         <version.commons.codec>1.13</version.commons.codec>
         <version.encoder>1.2.1</version.encoder>
-        <versions.xnio>3.8.5.Final</versions.xnio>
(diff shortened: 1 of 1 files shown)


stevehu:
I have updated all the repos with different branches. 

https://github.com/networknt/light-config-test/issues/655

PR #1168, opened 2022-03-17
Update on config loader for nested values.yml

wswjwjccjlu:


stevehu (approved):

stevehu:
@wswjwjccjlu Thanks a lot for your help. This is a very core function and very important when using the config server. I am wondering if you could add several test cases for map and list as well as string. When a list and map are used, indentation is very important as well. So we might need to test with different indentations.

PR #1194, opened 2022-03-29
fixes #1191 We have ProxyHandler in both egress-router and ingress-pr…

### LINKED ISSUE #1191: We have ProxyHandler in both egress-router and ingress-proxy at the same time
Need to compare them and see if we can share on in another module.

stevehu:
…oxy at the same time


GavinChenYan (approved):
looks good

PR #1197, opened 2022-03-30
Feature/content length error message

KalevGonvick:
- Added more checks to see if the payload is too large for undertow max buffer or if the payload has invalid formatting.


body/src/main/java/com/networknt/body/ProxyBodyHandler.java
```diff
@@ -98,14 +99,18 @@ public void handleRequest(final HttpServerExchange exchange) throws Exception {
...
                 Connectors.resetRequestChannel(exchange);
-            } catch (Exception | Error e) {
-                if (buffer != null && buffer.isOpen()) {
+            } catch (IOException e) {
+                if (buffer.isOpen()) {
                     IoUtils.safeClose(buffer);
                 }
-                throw e;
+                logger.error(e.getLocalizedMessage(), e);
+                return;
```
stevehu, line 107: I think we need to return a status object to indicate something is wrong. There is a standard exception status code in the status.yml with an error message as the parameter.

stevehu (changes requested):

Issue #1287, opened 2022-06-16
Decrypt encrypted values read from Environment


AkashWorkGit:
Issue:
Currently, when light-4j reads encrypted configuration values from the environment, it does not decrypt those values.

Proposed Feature/Update:
Decrypt encrypted values that are read from the environment.

stevehu:
This has been merged and tested.

Issue #1274, opened 2022-06-13
Environment Variable name look-up


AkashWorkGit:
Issue:
If light-4j is configured to pull a variable value from the environment, and the variable does not have a legal env-var name (e.g.: the token is my.server.ip), then light-4j will be unable to get any value for that variable.

Proposed Feature/Update:
In cases where a value comes from the environment, we want light-4j to convert the variable name to a legal env-var name before trying to find the value (e.g.: look for the env-var named MY_SERVER_IP instead of my.server.ip).

So this new feature will remove any alpha-numeric values from the naming convention and replace the "." with an "_" and then converting it to uppercase.

This behaviour is similar to functionality found in frameworks such as Spring.

stevehu:
I totally agree with you and we are in the process to convert some variables. The first step is to convert some secrets and passwords. Let us know what else we need to convert.

Here is an example we are doing. 

https://github.com/networknt/light-4j/blob/master/client/src/main/resources/config/client.yml#L27

miklish:
Hi @stevehu ,

Its good to see you start the process of updating the default variable names. :)

_However_ -- just to make sure we're on the same page -- we are not proposing changing the default variable names _in this issue_ (though we think that's a good idea). What we are proposing in this issue would actually allow light-4j to keep the existing variable names, but have light-4j dynamically change the names to a legal env-var format whenever it tries to find that value in the environment.

For example: This would allow light-4j to continue to use the variable name **client.keyStore** in the default config files... However, in the case that light-4j needs to get the value of **client.keyStore** from an environment variable, it would look for the value of **CLIENT_KEYSTORE** rather than **client.keyStore**. It would automatically do this name conversion (**client.keyStore** -> **CLIENT_KEYSTORE**) at runtime, and it would only do this name conversion when it attempts to get the value from the environment.

AkashWorkGit:
Hi Steve, Yes we are going to submit a PR for this to get implemented.

stevehu:
This issue has been completed on both 1.6.0 and 2.0.0 branches.

PR #1311, opened 2022-07-20
Changes made to the code as required. Mentioned in issue 1274 and iss…

AkashWorkGit:
Please review the code and let me know if there is anything else that I need to do before we merge it to the 1.6.x branch.


AkashWorkGit:
As mentioned in issue #1274 & #1287 and as discussed, with this new PR we give light-4j the capability of automatically converting the variable names to match the legal Environment variable format(Upper case and an underscore replacing the dot) at runtime. Also it will only do this conversion while it fetches the values from the Environment and would not affect the default variable naming convention that light-4j has as of now.

Also we are enhancing light-4j's capability to Decrypt encrypted values that are read from the environment.

**NOTE:** : Since We wanted to use the private method called getDecryptorClass() from the Config class, in our new method called getDecryptor() in the ConfigInjection class. We used it by creating a getDecryptorClassPublic() method in the Config class which returns the private getDecryptorClass() method and used it in the getDecryptor() method that we added in the ConfigInjection class.

PR #1323, opened 2022-08-08
Add a time buffer before terminate the blocking query connection to C…

jsu216:
…onsul


jsu216:
@stevehu , we decided to add a buffer before timeout since it is the same setting as the long blocking query to consul to avoid unnecessary timeout exception from the router. Please review and we would like a snapshot build ASAP. Thanks.

jaydeepparekh1311 (approved):

stevehu (approved):

PR #1340, opened 2022-08-15
Fix for 1339

KalevGonvick:
* Added new condition where content length is updated only when the response contains the content-length header already.
* Fixed typo


stevehu (approved):

stevehu:
@KalevGonvick Thanks for your help. I have released a new snapshot version for integration testing. Let me know if you have questions.

PR #1419, opened 2022-10-21
Fix flaky test in URLNormalizerTest.java

yannizhou05:
The test below in `URLNormalizerTest.java` was found flaky because of the ordering of result String. The orders of elements in the String are not the same every time being called. In this case, two possible outputs of the String are applied to ensure the orders do not cause the flakiness. This code change is trying to fix the [flaky tests](https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html) mentioned above, because they sometimes fail (as the picture showed) and sometimes pass. The failure could be reproduced by the commands above by using the tool of [NonDex](https://github.com/TestingResearchIllinois/NonDex). The code change is to make sure the tests will always pass in this case.

- `com.networknt.url.URLNormalizerTest.testAllAtOnce`

The test failures could be reproduced by

1. `mvn install -pl http-url -am -DskipTests`
2.  run tests with NonDex 
    `mvn -pl http-url edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=com.networknt.url.URLNormalizerTest#testAllAtOnce`

### DIFF
--- a/http-url/src/test/java/com/networknt/url/URLNormalizerTest.java
+++ b/http-url/src/test/java/com/networknt/url/URLNormalizerTest.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 public class URLNormalizerTest {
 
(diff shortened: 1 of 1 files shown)


stevehu (approved):

stevehu:
@yannizhou05 Thanks a lot for your help.

Issue #1498, opened 2022-11-24
Flaky Test In URLTest.java


tt0suzy:
`com.networknt.registry.URLTest.testURL`

This is a flaky test, it can pass mvn test while but when run using the tool [NonDex](https://github.com/TestingResearchIllinois/NonDex), it fails. NonDex is a tool that will introduce non-determinism in certain java collections.The test shows below:

```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.336 sec <<< FAILURE! - in com.networknt.registry.URLTest
testURL(com.networknt.registry.URLTest)  Time elapsed: 0.334 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<...value1&key2=true&key[5=10L&key3=10&key4=3.14]&nodeType=service&ve...> but was:<...value1&key2=true&key[3=10&key4=3.14&key5=10L]&nodeType=service&ve...>
        at com.networknt.registry.URLTest.testURL(URLTest.java:102)
```
[com.networknt.registry.URLTest.txt](https://github.com/networknt/light-4j/files/10080176/com.networknt.registry.URLTest.txt)

**Steps to reproduce the failure:**
 Run the following command in light-4j:
 - First, build the module:
 `mvn install -pl registry -am -DskipTests`
- Then run the test using NonDex 
`mvn -pl registry edu.illinois:nondex-maven-plugin:2.1:nondex -Dtest=com.networknt.registry.URLTest#testURL`
- The result will be saved under the module folder in .nondex

stevehu:
merged and closed.

PR #1499, opened 2022-11-24
Fix flaky test

tt0suzy:
[Issue regarding this PR](https://github.com/networknt/light-4j/issues/1498)
The issue is caused by HashMaps. The orders of elements in the HashMaps are not the same every time being called. And when you pass the `parameters` HashMap to `URLIMpl`, the order of parameters in the resulting URL is non-deterministic. In this case, LinkedHashMap is applied to replace Hash in order to ensure the orders are not changing.


stevehu:
@tt0suzy Thanks a lot for your help.

PR #1704, opened 2023-04-10
fixes #1703 integrate Http2Client with Simple Pool

### LINKED ISSUE #1703: integrate Http2Client with Simple Pool

stevehu:


miklish:
FYI: Line 341 of Http2Client:
```java
public void restore(URI uri, SimpleConnectionHolder.ConnectionToken token) {
    SimpleURIConnectionPool pool = pools.get(uri);
    if(pool != null) {
        pool.restore(token);
    }
}
```

The ConnectionToken contains the guaranteed correct URI. Leaving it to the user to provide the correct URI could lead to accidentally trying to return a connection to the wrong SimpleURIConnectionPool, which could cause a connection leak.

I would suggest the following instead:
```java
public void restore(SimpleConnectionHolder.ConnectionToken token) {
    SimpleURIConnectionPool pool = pools.get(token.uri());
    if(pool != null) {
        pool.restore(token);
    }
}
```

stevehu:
@miklish Thanks a lot for the review and recommendations. I am updating it, and another PR is coming for reviewing.

PR #1750, opened 2023-05-03
fixes #1749 remove environment from the common tags for metrics handlers

### LINKED ISSUE #1749: remove environment from the common tags and change gateway module to sidecar
As the environment tag is not available for all services. The address and port number can uniquely identify each environment. 
The handlers in the gateway module are for http-sidecar only; changing the module name and config file name will eliminate confusion for users.

stevehu:


stevehu:
Cancelled

PR #1789, opened 2023-06-14
fixes #1788 make ConnectionToken holder method public to allow the cl…

### LINKED ISSUE #1788: make ConnectionToken holder method public to allow the client to close the connection safely

stevehu:
…ient to close the connection safely


miklish:
hi @stevehu , I just noticed this change. By allowing users to modify connection states outside of the synchronization of the borrow() and restore() methods of SimpleURIConnectionPool, I believe this may allow inconsistent states in the connection pool to manifest in a multithreading environment.

I will need to look into this more.

Issue #1808, opened 2023-06-26
Reset Consul lookup Index after connection failure


zabooma:
If any connection failure occurs with connections to Consul, the Consul index for all services using the failed connection(s) must be reset to zero (0).

This  applies to [ConsulRegistry](https://github.com/networknt/light-4j/blob/1.6.45/consul/src/main/java/com/networknt/consul/ConsulRegistry.java) class

PR #1809, opened 2023-06-26
issue_1808 - Issue a non-blocking call so that the cache is refreshed without waiting for a timeout

zabooma:
issue #1808 and also:
Cleanup unused OptionMap initialization in ConsulClientImpl constructor and import.

### DIFF
--- a/client/src/main/java/com/networknt/client/Http2Client.java
+++ b/client/src/main/java/com/networknt/client/Http2Client.java
@@ -67,6 +67,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import static io.undertow.client.http.HttpClientProvider.DISABLE_HTTPS_ENDPOINT_IDENTIFICATION_PROPERTY;
 
 /**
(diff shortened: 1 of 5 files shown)


miklish:
Hi @zabooma -- I need to understand the additional changes in this PR.

Why are there changes to ConsulClientImpl and Http2Client?

For example, why was this removed from ConsulClientImpl?

```java
optionMap = isHttp2() ? OptionMap.create(UndertowOptions.ENABLE_HTTP2, true) : OptionMap.EMPTY;
```
?

From the PR description, the only change needed is a 1 line change in ConsulRegistry.

miklish:
@zabooma: @AkashWorkGit informed me that the changes to ConsulClientImpl were supposed to be in https://github.com/networknt/light-4j/pull/1821

zabooma:
@miklish : optionMap was identified as not used after all the changes we have made recently (switching to a different connection pool use). On one of the calls you asked to remove it on any of the outstanding PRs.

AkashWorkGit:
The Junit could not be created for the changes in consulCientImpl class after several attempts but the manual unit test & functional testing was completed successfully.

jaydeepparekh1311 (approved):

miklish (approved):

PR #1835, opened 2023-07-24
fixes #1825 Decryptor problem with AES-GCM

### LINKED ISSUE #1825: Decryptor problem with AES-GCM
Hi, I hope my message finds you well.

I'm currently having some problems when updating Ligh4J from version 2.1.9 to a newer one. I'm encrypting my passwords using AES-GSM(256), but I'm no longer able to decrypt them. I've checked the past PRs and I saw this one: https://github.com/networknt/light-4j/pull/1687
The regex to mark a string as a CRYPT changed and my passwords don't match it anymore (special characters are gone, for instance). As far as I know, AES-GCM doesn't use salt, but initialization vectors. However, it's true that I'm not an expert in cryptography. My question here is about the fact that we can't use our own decryptors as they were before. Should we change our encryption?

Thank you very much and best regards.

stevehu:


ForwardKeys (approved):
Thanks, this will help the backwards compatibility with customised decryptors

PR #1922, opened 2023-10-19
fixes #1921 pass the request body to the transformer only if it is a …

### LINKED ISSUE #1921: pass the request body to the transformer only if it is a string

stevehu:
…string


DiogoFKT (commented):
Looks good to me

PR #1930, opened 2023-10-24
Fixes #1908 - CRYPT regex too strict, only allows AES crypto alike strings

### LINKED ISSUE #1908: Regex for encryption too strict
Hi! We've seen the fix on [https://github.com/networknt/light-4j/blob/master/config/src/main/java/com/networknt/config/yml/YmlConstants.java](url) has been reverted, and the constant used for the regex to match encrypted passwords is now way more restrictive than it used to be.
It used to be like:
`public static final Pattern CRYPT_PATTERN = Pattern.compile("^CRYPT:([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$");`
And now it's like:
`public static final Pattern CRYPT_PATTERN = Pattern.compile("CRYPT:[a-zA-Z0-9:]+");`
The consequence is all the passwords already in place cannot be decrypted, as they don't match this new regex. This change was fixed in [https://github.com/networknt/light-4j/issues/1825](url) to solve the issue, but it has been reverted. Could it be that was reverted by mistake?
Many thanks in advance

ForwardKeys:
CRYPT regexp in YmlConstants has been adapted only to allow AES alike passwords, rendering all previous passwords no longer valid. This regex also invalidates the possibility of using customised Decryptors.
This small changes fixes #1908, allowing in the future to use any kind of Decryptor.

### DIFF
--- a/config/src/main/java/com/networknt/config/yml/YmlConstants.java
+++ b/config/src/main/java/com/networknt/config/yml/YmlConstants.java
@@ -6,6 +6,6 @@ import org.yaml.snakeyaml.nodes.Tag;
 
 public class YmlConstants {
 	public static final Tag CRYPT_TAG = new Tag(Tag.PREFIX + "crypt");
-	public static final Pattern CRYPT_PATTERN = Pattern.compile("CRYPT:[a-zA-Z0-9:=/+]+");
+	public static final Pattern CRYPT_PATTERN = Pattern.compile("^CRYPT:.*$");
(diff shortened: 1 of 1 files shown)


stevehu (approved):

stevehu:
@ForwardKeys Thanks for your help. I was thinking of adding more characters if required, but your solution is the ultimate one.

Issue #1905, opened 2023-10-05
SimplePool Version 2


miklish:
PR: https://github.com/networknt/light-4j/pull/1963

Numerous valuable updates to SimplePool have been identified to simplify its code and streamline its logic.

#### **_Breaking Changes_**
  * rename class `SimpleConnectionHolder` to `SimpleConnectionState`
  * rename class `SimpleClientConnection` to `SimpleUndertowConnection`
  * rename class `SimpleClientConnectionMaker` to `SimpleUndertowConnectionMaker`
  * remove the `reuseConnection()` method from the `SimpleConnectionMaker` interface

#### Refactorings
  * rename the `readAllConnectionHolders()` method to `applyAllConnectionStates()` in `SimpleURIConnectionPool`
  * rename the `readConnectionHolder()` to `applyConnectionState()` in `SimpleURIConnectionPool`
  * rename variable `allKnownConnections` to `allTrackedConnections` in `SimpleURIConnectionPool`
  * rename functional interface `RemoveFromAllKnownConnections` to `RemoveFromTrackedConnections` in `SimpleURIConnectionPool`
  * rename variable `knownConnections` to `trackedConnections` in `SimpleURIConnectionPool`

#### Documentation updates
  * document time units of all time-based method parameters
  * add vital docs on the expected behaviour of implementations of the `SimpleConnectionMaker` interface
  * use the word 'thread' instead of 'process'
  * document details of thrown exceptions
  * add significant additional documentation for `SimpleConnectionMaker`, `SimpleConnectionPool`, and `SimpleURIConnectionPool` classes
  * add docs explaining that `Set` variable `allCreatedConnections` must be threadsafe in `SimpleConnectionState`
  * add docs on how the `now` time variables are encoded as Unix Epoch time in milliseconds in `SimpleConnectionState`

#### General Code Changes
  * make `SimpleUndertowConnectionMaker` a threadsafe singleton
  * annotate interface `RemoveFromTrackedConnections` as a functional interface in `SimpleURIConnectionPool`
  * move the responsibility to close expired connections to the method `applyAllConnectionStates()` in `SimpleURIConnectionPool`
  * move the responsibility to close leaked connections from the method `applyAllConnectionStates()` to the `borrow()` and `restore()` methods of `SimpleURIConnectionPool`
  * update logic in `SimpleUndertowConnectionMaker` and `ThreadRunner` to handle removal of the `reuseConnection()` method from the `SimpleConnectionMaker` interface
[...]

PR #1963, opened 2023-11-08
SimplePool v2

miklish:
# SimpleConnectionPool v2
`SimpleConnectionPool` is a simple, natively mockable, threadsafe, resilient, high-performance connection pool intended to replace `com.networknt.client.http.Http2ClientConnectionPool`.

Use of this connection pool fully resolves issue 1656 (https://github.com/networknt/light-4j/issues/1656) by replacing the use of `Http2ClientConnection` in `com.networknt.consul.client.ConsulClientImpl.lookupHealthService()` with `SimpleURIConnectionPool`.

## Breaking Changes
  * class `SimpleConnectionHolder` renamed to `SimpleConnectionState`
  * class `SimpleClientConnection` renamed to `SimpleUndertowConnection`
  * class `SimpleClientConnectionMaker` renamed to `SimpleUndertowConnectionMaker`
  * method `reuseConnection()` removed from the `SimpleConnectionMaker` interface

*See issue https://github.com/networknt/light-4j/issues/1905 for more details on changes.*

## Focus on Stability and Testability
The primary goal of SimpleConnectionPool is operational stability and reliability. Optimizations were considered only after stability and resilience were confirmed.

Easy testability was vital to confirm the correct behaviour of the connection pool under a wide array of common and corner-case conditions. To this end, the connection pool was developed with mockability built-in.

## Avoidance of Consul Connection Bug

The following Consul connection-leak bug was reported in August 2020, but has yet to be resolved.

> When a Consul client cancels a blocking HTTP-query, the TCP connection is not closed correctly by the server. The TCP connection stays in the FIN_WAIT-2 state until it's tcp_fin_timeout expires. The FIN_WAIT-2 means that the client is waiting for an ACK from the server.
> 
> When a lot of blocking queries are cancelled and retried at the same time, the servers http_max_conns_per_client can be hit the further client queries will fail.

> Bug Ticket URL: https://github.com/hashicorp/consul/issues/8524

SimpleConnectionPool avoids this bug entirely by never closing connections that are in use. While this can cause connection expiry times to not be precisely honoured, it also ensures that Consul blocking queries are never canceled before they complete, thereby avoiding the conditions that manifest this bug.

## Multi-Layer Architecture
The connection pool was developed using a multilayer architecure.

### Level 1: SimpleConnectionPool routes requests to URI connection pools
Threads that need a connection to a URI must 'borrow' a connection from the pool and then `restore` that connection to the pool when they are done with it.

A `SimpleConnectionPool` routes these URI-connection requests to a `SimpleURIConnectionPool` for that URI. If a `SimpleURIConnectionPool` does not exist for that URI, the `SimpleConnectionPool` will create one.

The `SimpleConnectionPool` needs only minimal thread synchronization since the `SimpleURIConnectionPool`s are already fully thread safe. A benefit of this, is increased opportunity for concurrency. For example: N threads can request connections to N distinct URIs concurrently.

### Level 2: SimpleURIConnectionPools manage connections to a single URI
A `SimpleURIConnectionPool` manages connections to a single URI where the connections it manages have the ability to be used by 1 or more threads concurrently. A `SimpleURIConnectionPool`:

- Enforces a configurable maximum number of connections
- Closes connections after a configurable 'expiry' time, but ensures that it never closes connections that are in use
- Safely manages connection resources (if client behaves as expected)
- Safely closes leaked connections that may be created by connection-creation callback threads after a connection-creation timeout has occurred in the parent thread (more on this below)
- Is threadsafe

Also, note that `SimpleURIConnectionPool`s can be used used independently of `SimpleConnectionPool`. This means that, any code that only needs to connect to a single URI can use a `SimpleURIConnectionPool` directly--it does not need to make requests to borrow connections from a `SimpleConnectionPool` (which is only needed for code that needs to connect to multiple distinct URIs).

The `SimpleURIConnectionPool` needs very little information about connections to manage the pool. It only needs to know if the connection is:

- **borrowed**
- **borrowable**
- **expired**
- **closed**

Note: More than one of these properties can simultaneously apply to a connection (e.g.: an HTTP/2 connection may be both **borrowed** and **borrowable**).

### Level 3: SimpleConnectionStates manage connection states

`SimpleConnectionState` objects both create and manage the state of a single network connection.

`SimpleConnectionState`s have a 1-1 relationship with their connection. In other words:

- every connection in the pool is created by -- and contained in -- a single `SimpleConnectionState`, and
- every `SimpleConnectionState` contains the single connection it created.

When a `SimpleURIConnectionPool` needs a new connection, it will create a new `SimpleConnectionState` object. The connection pool will then manage that connection exclusively by communicating through the connection's enclosing `SimpleConnectionState`, rather than managing the connection directly.

The `SimpleConnectionState` provides the following information about a connection to the pool:

- **borrowed**
  - A connection is considered 'borrowed' if 1 or more threads are currently using the connection at the moment.
- **borrowable**
  - A connection is considered 'borrowable' if the connection can be used by a new thread
    - _HTTP/1.1_: HTTP/1.1 connections are only borrowable when 0 connection tokens are currently borrowed
    - _HTTP/2_: HTTP/2 connections are  always be borrowable
- **expired** / **valid**
  - A connection is 'expired' once its age exceeds the pool's connection expiry time setting
  - A connection is 'valid' if it is not yet expired
- **closed**
  - A connection is closed if the the pool or the OS closes it

The state diagram of a `SimpleConnectionState` is shown below.

#### SimpleConnectionState - State diagram for a connection
 <pre>
              |
             \/
    [ NOT_BORROWED_VALID ]  --(borrow)--&gt;  [ BORROWED_VALID ]
              |             &lt;-(restore)--          |
              |                                    |
           (expire)                             (expire)
              |                                    |
             \/                                   \/
    [ NOT_BORROWED_EXPIRED ] &lt;-(restore)-- [ BORROWED_EXPIRED ]
             |
          (close) (*)
             |
            \/
        [ CLOSED ]
 </pre>

> (*) ***NOTE***<br/>
> A connection can be closed explicitly by the connection pool, or it can be unexpectedly closed at any time by the OS. If it is closed unexpectedly by the OS, then the state can jump directly to CLOSED regardless of what state it is currently in.

### Level 4: SimpleConnections and SimpleConnectionMakers

`SimpleConnection` interfaces wrap physical 'raw' network connections. For example, the `SimpleUndertowConnection` class implements `SimpleConnection` and wraps Undertow's `ClientConnection`.

`SimpleConnectionMaker`'s are factories that create `SimpleConnection` objects, and are used by `SimpleConnectionState` objects to create the connection they manage. Different implementations of `SimpleConnectionMaker` can be used to create different types of raw connections. For example: A `SimpleUndertowConnectionMaker` uses the Undertow networking library to create connections, a `SimpleApacheConnectionMaker` could use the Apache libraries, and a `SimpleMockConnectionMaker` could create mock connections used for testing.

_Note:_ `SimpleConnectionState`s only deal with `SimpleConnection` objects -- they never deal directly with 'raw' connections (such as Undertow's `ClientConnection` objects).

#### Creating a new network (raw) connection
1. Create a `SimpleConnectionMaker` (e.g.: a `SimpleUndertowConnectionMaker`)
<pre>
[Connection Pool]   [Connection State]   [Connection Maker]   [Simple Connection]   [Raw Connection]
       [|]                   |                    |                    |                    |
       [|]   &lt;&lt; creates &gt;&gt;   |                    |                    |                    |
       [|]-------------------------------------->[|]                   |                    |
       [|]                   |                   [|]                   |                    |
</pre>
<br/>

2. Pool creates new connections by instantiating new `SimpleConnectState` objects (passing them a `SimpleConnectionMaker` in their constructor)
3. `SimpleConnectionState` uses the `SimpleConnectionMaker` to create `SimpleConnection` objects that wrap raw network connections
<pre>
[Connection Pool]   [Connection State]   [Connection Maker]   [Simple Connection]   [Raw Connection]
       [|]                   |                   [|]                   |                    |
       [|]   &lt;&lt; creates &gt;&gt;   |                   [|]                   |                    |
       [|]----------------->[|]                  [|]                   |                    |
       [|]                  [|]    &lt;&lt; uses &gt;&gt;    [|]                   |                    |
       [|]                  [|]----------------->[|]                   |                    |
       [|]                  [|]                  [|]   &lt;&lt; creates &gt;&gt;   |                    |
       [|]                  [|]                  [|]----------------->[|]                   |
       [|]                  [|]                  [|]                  [|]   &lt;&lt; creates &gt;&gt;   |
       [|]                  [|]                  [|]                  [|]----------------->[|]
</pre>

## Finding and closing leaked connections
Typically, new connections are created in a callback thread spawned by the thread that will use the connection.

### Example of connection leak detection: SimpleUndertowConnectionMaker

To create a new connection, a `SimpleUndertowConnectionMaker` will spawn a new thread to create the connection. It will then wait a configurable amount of time (a timeout) for the connection-creation thread to finish creating the connection.

#### Case 1: Connection creation failed (OK)

If the connection-creation thread is unable to create the connection, then `SimpleUndertowConnectionMaker` will throw a 'connection-creation' exception and `SimpleURIConnectionPool` will not recieve a new connection.

#### Case 2: Connection successfully created before timeout (OK)

If the connection is successfully created before the timeout expires, then `SimpleUndertowConnectionMaker` will get the new connection from the connection-creation thread and return it to `SimpleConnectionState` (and, in turn, the new `SimpleConnectionState` holding the new connection will be returned to the pool).

#### Case 3: Connection successfully created *after* timeout (!)

If the connection is successfully created ***after*** the timeout expires, then -- as in case 1 -- `SimpleUndertowConnectionMaker` will throw a 'connection-creation' exception and `SimpleURIConnectionPool` will not receive a new connection.

However, in contrast to case 2: since a new connection was successfully created but not returned to the pool, it means that this new connection will exist but be unknown to the `SimpleURIConnectionPool`. Without knowledge of the connection, `SimpleURIConnectionPool` cannot not close it and so this `untracked` connection is considered a **connection leak**.

### Handling connections created after timeouts

Case 3 results in connections that have been created by a `SimpleConnectionMaker` but are unknown to the pool and  are called 'untracked' connections (conversely, connections known to the pool are called 'tracked' connections).

To ensure that all untracked connections are closed, users of a `SimpleConnectionMaker` must pass a threadsafe `Set<SimpleConnection>` to the `SimpleConnectionMaker`'s `makeConnection()` method. Implementations of `SimpleConnectionMaker` must add all connections they create to this `Set`.

To ensure all untracked connections are closed, `SimpleURIConnectionPool` will periodically remove (but not close) all tracked connections from this set (e.g.: it will remove the connections it knows about). It will then close close any connections that remain in the set, since these remaining connections are untracked. This ensures that any untracked connections in the `Set` are safely closed. 

## Mockability
Since `SimpleConnectState` objects use a `SimpleConnectionMaker` to create their connections, and since the connections that `SimpleConnectionMakers` create implement the `SimpleConnection` interface, it means that `SimpleURIConnectionPool`, and `SimpleConnectionState` are able to manage the pool without having any knowledge of the raw connections they are managing.

This means mock connections can be created by simply implementing `SimpleConnection` and creating a `SimpleConnectionMaker` that instantiates these mock connections.

#### Mock Test Harness
SimpleConnectionPool comes with a test harness that can be used to easily test mock connections. For example, to test how the connection pool will handle a connection that randomly closes can be built as follows:

**1. Develop the mock connection**
```java
public class MockRandomlyClosingConnection implements SimpleConnection {
    private volatile boolean closed = false;
    private boolean isHttp2 = true;
    private String MOCK_ADDRESS = "MOCK_HOST_IP:" + ThreadLocalRandom.current().nextInt((int) (Math.pow(2, 15) - 1.0), (int) (Math.pow(2, 16) - 1.0));

    /***
     * This mock connection simulates a multiplexable connection that has a 5% chance of closing
     *  every time isOpen() is called
     */
    public MockRandomlyClosingConnection(boolean isHttp2) { this.isHttp2 = isHttp2; }
    @Override public boolean isOpen() {
        if(ThreadLocalRandom.current().nextInt(20) == 0)
            closed = true;
        return !closed;
    }
    @Override public Object getRawConnection() {
        throw new RuntimeException("Mock connection has no raw connection");
    }
    @Override public boolean isMultiplexingSupported() { return isHttp2; }
    @Override public String getLocalAddress() { return MOCK_ADDRESS; }
    @Override public void safeClose() { closed = true; }
}
```

**2. Test how the connection pool handles the connection**
```java
public class TestRandomlyClosingConnection {
    public static void main(String[] args) {
        new TestRunner()
            // set connection properties
            .setConnectionPoolSize(100)
            .setSimpleConnectionClass(MockRandomlyClosingConnection.class)
            .setCreateConnectionTimeout(5)
            .setConnectionExpireTime(5)
            .setHttp2(true)

            // configure borrower-thread properties
            .setNumBorrowerThreads(8)
            .setBorrowerThreadStartJitter(3)
            .setBorrowTimeLength(5)
            .setBorrowTimeLengthJitter(5)
            .setWaitTimeBeforeReborrow(2)
            .setWaitTimeBeforeReborrowJitter(2)

            // execute test
            .setTestLength(10*60)
            .executeTest();
    }
}
```

## Plugability
`SimpleConnectionPool` requires very litte information about a connection in order to manage that connection. This lends itself to supporting many different networking APIs. If one can implement a `SimpleConnection` and `SimpleConnectionMaker` using a networking API, then connections created using that API can be managed by `SimpleConnectionPool`.

`ClientConnections` created using undertow's libraries can be wrapped this way. See the `simplepool.undertow` package to see how support for undertow was implemented.

## How to safely use SimpleConnectionPool and SimpleURIConnectionPool
The following code snippet demonstrates the recommended way to structure code that borrows a connection.

*Note: The code below is the same whether <code>pool</code> is a SimpleConnectionPool or a SimpleURIConnectionPool*.

```java
SimpleConnectionState.ConnectionToken borrowToken = null;
try {
    borrowToken = pool.borrow(createConnectionTimeout, isHttp2);
    ClientConnection connection = (ClientConnection) borrowToken.getRawConnection();

    // Use connection...
    
} finally {
    // restore token
    pool.restore(borrowToken);
}    
```

...and this demonstrates the recommended way to borrow connections in a loop...

```java
while(true) {
    SimpleConnectionState.ConnectionToken borrowToken = null;
    try {
        borrowToken = pool.borrow(createConnectionTimeout, isHttp2);
        ClientConnection connection = (ClientConnection) borrowToken.getRawConnection();

        // Use connection...
        
    } finally {
        // restore token
        pool.restore(borrowToken);
    }    
}
```

### DIFF
--- a/client/src/main/java/com/networknt/client/oauth/OauthHelper.java
+++ b/client/src/main/java/com/networknt/client/oauth/OauthHelper.java
@@ -20,9 +20,9 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import com.networknt.client.ClientConfig;
 import com.networknt.client.Http2Client;
-import com.networknt.client.simplepool.SimpleConnectionHolder;
+import com.networknt.client.simplepool.SimpleConnectionState;
 import com.networknt.client.simplepool.SimpleConnectionPool;
-import com.networknt.client.simplepool.undertow.SimpleClientConnectionMaker;
+import com.networknt.client.simplepool.undertow.SimpleUndertowConnectionMaker;
 import com.networknt.cluster.Cluster;
 import com.networknt.config.Config;
 import com.networknt.exception.ClientException;
@@ -88,7 +88,7 @@ public class OauthHelper {
     private static final Logger logger = LoggerFactory.getLogger(OauthHelper.class);
 
     private static final SimpleConnectionPool pool = new SimpleConnectionPool(
-            ClientConfig.get().getConnectionExpireTime(), ClientConfig.get().getConnectionPoolSize(), SimpleClientConnectionMaker.instance());
+            ClientConfig.get().getConnectionExpireTime(), ClientConfig.get().getConnectionPoolSize(), SimpleUndertowConnectionMaker.instance());
 
     /**
      * @deprecated As of release 1.5.29, replaced with @link #getTokenResult(TokenRequest tokenRequest)
@@ -413,7 +413,7 @@ public class OauthHelper {
         final Http2Client client = Http2Client.getInstance();
         final CountDownLatch latch = new CountDownLatch(1);
         ClientConnection connection = null;
-        SimpleConnectionHolder.ConnectionToken borrowToken = null;
+        SimpleConnectionState.ConnectionToken borrowToken = null;
 
         long connectionTimeout = Math.max(2, keyRequest.getKeyConnectionTimeout() / 1000);
         long populateKeyTimeout = Math.max(2, keyRequest.getPopulateKeyTimeout() / 1000);
--- a/client/src/main/java/com/networknt/client/simplepool/SimpleConnection.java
+++ b/client/src/main/java/com/networknt/client/simplepool/SimpleConnection.java
@@ -22,7 +22,7 @@ package com.networknt.client.simplepool;
 
 /***
  * SimpleConnection is an interface that contains all the required functions and properties of
- * a connection that are needed by the SimpleConnectionHolder, SimpleURIConnectionPool, and
+ * a connection that are needed by the SimpleConnectionState, SimpleURIConnectionPool, and
  * SimpleConnectionPool classes.
  *
  * Concrete HTTP network connections (like Undertow's ClientConnection class) should be wrapped in
--- a/client/src/main/java/com/networknt/client/simplepool/SimpleConnectionMaker.java
+++ b/client/src/main/java/com/networknt/client/simplepool/SimpleConnectionMaker.java
@@ -24,10 +24,46 @@ import java.util.Set;
 
 /***
  * A factory that creates raw connections and wraps them in SimpleConnection objects.
- * SimpleConnectionMakers are used by SimpleConnectionHolders to create connections.
+ * SimpleConnectionMakers are used by SimpleConnectionStates to create connections.
  *
  */
 public interface SimpleConnectionMaker {
-    public SimpleConnection makeConnection(long createConnectionTimeout, boolean isHttp2, final URI uri, final Set<SimpleConnection> allCreatedConnections);
-    public SimpleConnection reuseConnection(long createConnectionTimeout, SimpleConnection connection) throws RuntimeException;
+    /***
+     * Establishes a new connection to a URI.
+     * Implementations of SimpleConnectionMaker are used by SimpleConnectionStates as a connection factory.
+     *
+     * @param createConnectionTimeout the maximum time in seconds to wait for a connection to be established
+     * @param isHttp2 if true, SimpleConnectionMaker must attempt to establish an HTTP/2 connection, otherwise it will
+     *          attempt to create an HTTP/1.1 connection
+     * @param uri the URI to connect to
+     * @param allCreatedConnections Implementations of SimpleConnectionMaker are used by SimpleConnectionState to create
+     *          connections to arbitrary URIs. In other words, implementations of SimpleConnectionMaker are used by
+     *          SimpleConnectionState as a connection factory.
+     *
+     *          A SimpleConnectionMaker will add all connections it creates to the Set <code>allCreatedConnections</code>.
+     *          SimpleURIConnectionPool will compare the connections in this Set to those that it is tracking and close
+     *          any untracked connections.
+     *
+     *          Untracked connections can occur if there is a connection creation timeout. When such a timeout occurs,
+     *          makeConnection() must throw a RuntimeException which will prevent SimpleURIConnectionPool from acquiring
+     *          a SimpleConnection. However, the connection creation callback thread in makeConnection() may continue to
+     *          execute after the timeout and ultimately succeed in creating the connection after the timeout has occurred
+     *          and the exception has been thrown. Connections that are created but were not returned to
+     *          SimpleURIConnectionPool are considered to be 'untracked'.
+     *
+     *          Despite not being tracked by SimpleURIConnectionPool, all successfully created connections must be added
+     *          to <code>allCreatedConnections</code>.
+     *
+     *          SimpleURIConnectionPool prevents these untracked connections from accumulating and causing a connection
+     *          leak over time, by periodically closing any open connections in allCreatedConnections that it is not tracking.
+     *
+     *          Thread Safety:
+     *              allCreatedConnections MUST be a threadsafe Set, or the thread safety of the connection pool cannot
+     *              be guaranteed.
+     *
+     * @return A SimpleConnection to the specified URI
+     * @throws RuntimeException thrown if the connection establishment timeout (<code>createConnectionTimeout</code>)
+     *          expires before a connection to the URI is established, or if there is an error establishing the connection
+     */
+    public SimpleConnection makeConnection(long createConnectionTimeout, boolean isHttp2, final URI uri, final Set<SimpleConnection> allCreatedConnections) throws RuntimeException;
 }
--- a/client/src/main/java/com/networknt/client/simplepool/SimpleConnectionPool.java
+++ b/client/src/main/java/com/networknt/client/simplepool/SimpleConnectionPool.java
@@ -24,7 +24,11 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
- * This is a class through which multiple URI Connection Pools can be accessed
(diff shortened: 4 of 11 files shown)


AkashWorkGit (approved):

jaydeepparekh1311 (approved):

miklish:
Hi @stevehu , we hope you can review and merge this on Tuesday. Thanks!

miklish:
Hi @stevehu ,

If possible, please merge this at your earliest convenience, or please let us know if you have any concerns about this PR so we can discuss and address them.

thanks!

PR #856, opened 2020-12-08
Fixed flaky tests caused by SharedMetricRegistries in SharedMetricRegistriesTest.java

KellyShao:
Fixed flaky test that caused by `io.dropwizard.metrics.SharedMetricRegistries` in class `SharedMetricRegistriesTest.java`. 

- **Issue 1** 
    - If only run `errorsWhenDefaultAlreadySet()` method before `hasASetOfNames()`method, the `hasASetOfNames()` will failed at https://github.com/networknt/light-4j/blob/fcded1683dcbd41a968e221494778aa6b71e7428/metrics/src/test/java/io/dropwizard/metrics/SharedMetricRegistriesTest.java#L45

    - error msg: 
    ```
        java.lang.AssertionError: 
        Expecting:
          <["one", "borg", "foobah"]>
        to contain only:
          <["one"]>
        but the following elements were unexpected:
          <["borg", "foobah"]>
    ```

- **Issue 2** 
    - If only run `errorsWhenDefaultAlreadySet()` method before `removesRegistries()`method, the `removesRegistries()` will failed at https://github.com/networknt/light-4j/blob/fcded1683dcbd41a968e221494778aa6b71e7428/metrics/src/test/java/io/dropwizard/metrics/SharedMetricRegistriesTest.java#L54

    - error msg: 
    ```
       java.lang.AssertionError: 
       Expecting:
         <["one", "foobah"]>
       to contain only:
         <["one"]>
       but the following elements were unexpected:
         <["foobah"]>
    ```
- **root cause**
    - The `SharedMetricRegistries` variable is not cleared after each tests. 

- **solution**: 
Added `setUp()` and `tearDown()` methods right before and after each tests to clear `SharedMetricRegistries`.

The flaky tests are found by running [iDFlakies](https://github.com/idflakies/iDFlakies), after fixing the issue, there's no flaky tests in `SharedMetricRegistriesTest.java`

PR #1996, opened 2023-11-14
Fix order dependent test in handler test

### LINKED ISSUE #2005: Unit test `validConfig_init_handlersCreated` in `HandlerTest` is order dependent
#### Issue
- If `validConfig_init_handlersCreated()` is run after `invalidMethod_init_throws()`, we face the following issue
```
java.lang.RuntimeException: Bad paths element in invalid-method config [ Invalid HTTP method: hello ]
	at com.networknt.handler.config.PathChain.validate(PathChain.java:94)
	at com.networknt.handler.Handler.initPaths(Handler.java:127)
	at com.networknt.handler.Handler.init(Handler.java:75)
	at com.networknt.handler.HandlerTest.test02_validConfig_init_handlersCreated(HandlerTest.java:57)
```
- Since JUnit 4 doesn't guarantee the order of test executions by default, JUnit could execute `validConfig_init_handlersCreated()` after `invalidMethod_init_throws()` resulting in the above error.

#### Root Cause
- The test `invalidMethod_init_throws()` sets the `HandlerConfig` to `invalid-method` at https://github.com/networknt/light-4j/blob/7152f180fe4eef1e991b5618d61ca199c3d0de77/handler/src/test/java/com/networknt/handler/HandlerTest.java#L121
- When the test `validConfig_init_handlersCreated()` is run after the above step, the `HandlerConfig` is no longer the default value `handler`
- Initiating the `HandlerTest.validConfig_init_handlersCreated()` with `invalid-method` configuration causes the above issue. 

#### Proposed Solution
- Resetting the `HandlerConfig` to default value before each test run will ensure that the unit tests are run with the correct `HandlerConfig`
- This can be done by using the Junit's `@Before` annotation, where the configuration is set to default value through `Handler.setConfig('handler')`

SaaiVenkat:
#### Issue
- In `com.networknt.handler.HandlerTest`, the unit test `validConfig_init_handlersCreated()` can trigger `java.lang.RuntimeException` when being run before the unit test `invalidMethod_init_throws()` in the same class because it pollutes state shared among tests.
- It may be good to clean this state pollution so that some other tests do not fail in the future due to the shared state polluted by this test.

Fixes #2005 

#### Solution
- Resetting the `HandlerConfig` to default value before each test run using the Junit's `@Before` annotation, where the configuration is set to default value through `Handler.setConfig('handler')`
- Adding `initChain()` because `initPath()`  in `setConfig`, which expects handler or chain to already be created.

### DIFF
--- a/handler/src/main/java/com/networknt/handler/Handler.java
+++ b/handler/src/main/java/com/networknt/handler/Handler.java
@@ -540,6 +540,7 @@ public class Handler {
         Handler.configName = configName;
         config = (HandlerConfig) Config.getInstance().getJsonObjectConfig(configName, HandlerConfig.class);
         initHandlers();
+        initChains();
         initPaths();
     }
 
--- a/handler/src/test/java/com/networknt/handler/HandlerTest.java
+++ b/handler/src/test/java/com/networknt/handler/HandlerTest.java
@@ -24,6 +24,7 @@ import io.undertow.util.HttpString;
 import io.undertow.util.Methods;
 import io.undertow.util.PathTemplateMatcher;
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 
 import java.util.Arrays;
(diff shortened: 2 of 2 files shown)


SaaiVenkat:
Greetings @stevehu, similar to #856, this PR also fixes the order dependency issue. I am happy to hear your feedback and discuss more about this issue.

stevehu (approved):

stevehu:
@SaaiVenkat It makes sense. Thanks a lot for your help.

PR #2002, opened 2023-11-17
Explicitly declare the used but undeclared dependency.

### LINKED ISSUE #2001: Undeclared dependency identified in client module.
# Problem desciption
The module client currently relies on a transitive dependency that hasn't been explicitly declared.

It seems that a wide range of classes of xnio-api are used by client both at compile and test time.
```
[org.xnio.ChannelListeners, org.xnio.IoFuture, org.xnio.ssl.SslConnection, org.xnio.ssl.XnioSsl, org.xnio.Xnio, org.xnio.XnioWorker, org.xnio.ChannelListener, org.xnio.conduits.ConduitStreamSinkChannel, org.xnio.ChannelExceptionHandler, org.xnio.channels.StreamSourceChannel, org.xnio.conduits.ConduitStreamSourceChannel, org.xnio.XnioIoThread, org.xnio.OptionMap, org.xnio.Options, org.xnio.StreamConnection, org.xnio.FutureResult, org.xnio.channels.StreamSinkChannel, org.xnio.conduits.PushBackStreamSourceConduit, org.xnio.IoUtils]
```
It runs correctly at present due to the introduction of this dependency through transitivity.
```
com.networknt:client:2.1.29-SNAPSHOT:: -> io.undertow:undertow-core:2.3.10.Final::compile -> org.jboss.xnio:xnio-api:3.8.8.Final::compile
```

# Possible Outcome:

Relying solely on transitive dependencies makes the project susceptible to changes outside your control. Failure or instability in the project's functionality due to changes in transitive dependencies caused by alterations or removal of the introducing transitive dependency.

# Possible Solution
We should explicitly declare the dependency in our pom.xml to ensure it is clearly documented as a project dependency.

HappyHacker123:
Fix #2001 .

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -88,6 +88,10 @@
             <artifactId>undertow-core</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.jboss.xnio</groupId>
+            <artifactId>xnio-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.jaegertracing</groupId>
             <artifactId>jaeger-client</artifactId>
             <optional>true</optional>
--- a/pom.xml
+++ b/pom.xml
@@ -98,6 +98,7 @@
         <version.jaeger>1.8.1</version.jaeger>
         <version.thrift>0.18.1</version.thrift>
         <version.opentracing>0.33.0</version.opentracing>
+        <version.xnioapi>3.8.8.Final</version.xnioapi>
         <version.org.apache.directory.server>2.0.0-M15</version.org.apache.directory.server>
         <version.maven-version>2.7</version.maven-version>
         <version.nexus-staging-maven>1.6.8</version.nexus-staging-maven>
@@ -577,6 +578,11 @@
                 <groupId>org.jboss.threads</groupId>
                 <artifactId>jboss-threads</artifactId>
                 <version>2.3.2.Final</version>
+            </dependency>
+            <dependency>
+                <groupId>org.jboss.xnio</groupId>
+                <artifactId>xnio-api</artifactId>
+                <version>${version.xnioapi}</version>
             </dependency>
             <dependency>
                 <groupId>org.hdrhistogram</groupId>


HappyHacker123:
@stevehu Could you please take a look at this pr? Thx :)

stevehu:
This is a transitive dependency from the Undertow. Do you have a reason to add that explicitly?

HappyHacker123:
Thanks for your quick reponse :). 

The problem mainly lies in that light-4j use the classes in xnio-api directly as mentioned in #2001. This means that light-4j relys on xnio-api **itself directly**. It's certain that you don't need to declare all the transitive dependency used by your directly dependency because maven will do the resolution, but xnio-api here is actually your direct dependency.

 If you uses the transitive dependnecy pulled in by undertow directly without declaring it, maven won't even know that your project relys on xnio-api, thus losing full control over your project's dependency. Let's assume the extreme cases (though it may not happen), if undertow abandons the use of xnio-api, light-4j won't be able to build. Or maybe undertow use upgraded xnio-api that removes certain classes, same thing will happen. The possible outcomes are induced by the used but undeclared dependency.

Actually, maven itself takes the act of using transitive dependency directly without declared as inappropriate. This can be proved by its official dependency managment plugin. You can take a look at https://maven.apache.org/plugins/maven-dependency-plugin/analyze-only-mojo.html if you are interested.  Hope that my reasons are sufficiently convincing, thanks for your patience :) @stevehu .

stevehu:
@HappyHacker123 I understand your concerns. We didn't put the xnio into the dependency because that is the hard dependency from Undertow, and it won't be changed in the feature confirmed by RedHat. By leaving xnio out of our dependency, we don't need to manage that library upgrade. At this moment, I think it

HappyHacker123:
> We didn't put the xnio into the dependency because that is the hard dependency from Undertow, and it won't be changed in the feature confirmed by RedHat.

@stevehu Thanks for your clarification :). I can see that optimizing the dependency structure may potentially introduce additional maintenance labor. So if it's confirmed by RedHat, then i understand your decision.

stevehu:
It is hard to draw the line for managing dependencies. If we include all transitive dependencies, we must add at least 30 more. So we must find a threshold between too few and too many. It was a good discussion and I never got a chance to think about it in depth. Thanks.

stevehu:
@HappyHacker123 Based on our conversation, I am closing this PR. Thanks.

Issue #2021, opened 2023-11-27
Add maven wrapper for light-4j.


HappyHacker123:
# Problem Description
Currently the project has not adapted maven wrapper, as the necessary Maven wrapper files, namely maven-wrapper.jar and maven-wrapper.properties ..., are not found.

The Maven Wrapper is an officially recommended way to ensure users of your Maven build has everything necessary to run your Maven build according to https://maven.apache.org/wrapper/.

# Benefits of Adding Maven Wrapper

With a Maven Wrapper setup, users can simply run wrapper scripts like this:
>    ./mvnw clean install

Even if maven is not installed or maven version is incompatible with light-4j setting. "./mvnw" can function exactly as "mvn" do.

Contributors can easily start working on the project without worrying about installing a specific version of Maven, but simply run the maven wrapper script.
 
 What's more, builds become reproducible across various environments, ensuring that everyone working on the project uses the same Maven version.

# Solution

Simply run the below command in the root project. Automatically the wrapper scripts will be generated by maven.  

>    mvn wrapper:wrapper

HappyHacker123:
I raise a pr to show how maven wrapper works. You can generate the proper setting suitable for ligh4-4j :).

PR #2022, opened 2023-11-27
Add maven wrapper.

HappyHacker123:
Related to #2021. 
This is just an example of automatically generated maven wrapper by the official maven plugin maven-wrapper-plugin.


stevehu:
@HappyHacker123 Thanks for your help. We generate new projects with the maven wrapper, which is very convenient for users.

Issue #2028, opened 2023-12-05
Ensure leaked connections are closed in SimpleURIConnectionPool.restore() (SimplePool v2) (1.6.x)


miklish:
PR: https://github.com/networknt/light-4j/pull/2042

(related to Issue: https://github.com/networknt/light-4j/issues/2053 - was intended to be merged before https://github.com/networknt/light-4j/issues/2053 - this is why the 'Files changed' is only a change to documentation)

In SimplePool v1, a leaked-connection check (`findAndCloseLeakedConnections()` method) was done in both the `borrow()` and `restore()` methods of `SimpleURIConnectionPool`. In SimplePool v2, the leaked-connection check in `restore()` was removed. This check should be added back.

The `findAndCloseLeakedConnections()` method must also be moved to the start of `borrow()`. When at the end of the `borrow()` method, the connection leak check can be skipped due to `borrow()` exiting via an exception (which can be thrown by `SimpleConnectionMaker.makeConnection()` if connection-creation fails) before executing the leak check.

The `SimpleConnectionState.borrow()` method needs to remove the unused `connectionCreateTimeout` argument. Since this method is only used internally by `SimpleURIConnectionPool`, it is not a breaking change.

Also, some refactorings and documentation updates can be done to clarify the code. These refactorings are at the SimplePool internal implementation level, and so are not breaking changes.

Finally, to coincide with re-addition of connection leak checking in the `restore()` method, and the change in its placement in both `borrow()` and `restore()`, the connection leak checking mock test has been updated (specifically, the `MockTimeoutLeakedConnection` class has been updated) to provide more information on connection leak finding and fixing.

stevehu:
* Optimized Leak Detection: Connection leak checks are now performed at the very start of both borrow() and restore() operations in SimpleURIConnectionPool, ensuring that any leaked resources are cleaned up immediately.
* API Cleanup: Removed the unused createConnectionTimeout parameter from SimpleConnectionMaker.reuseConnection() and SimpleConnectionHolder.borrow(), making the internal API cleaner and more consistent with the 1.6.x branch.
* Full Verification: Verified the changes by compiling the client module and running the SimpleURIConnectionPoolTest suite, which passed successfully.

PR #2042, opened 2023-12-12
Ensure leaked connections are closed in SimpleURIConnectionPool.restore() (SimplePool v2) (1.6.x)

miklish:
Issue: https://github.com/networknt/light-4j/issues/2028

(related to Issue: #2053 - was intended to be merged before #2053 - this is why the 'Files changed' is only a change to documentation)

In SimplePool v1, a leaked-connection check (`findAndCloseLeakedConnections()` method) was done in both the `borrow()` and `restore()` methods of `SimpleURIConnectionPool`. In SimplePool v2, the leaked-connection check in `restore()` was removed. This check should be added back.

The `findAndCloseLeakedConnections()` method must also be moved to the start of `borrow()`. When at the end of the `borrow()` method, the connection leak check can be skipped due to `borrow()` exiting via an exception (which can be thrown by `SimpleConnectionMaker.makeConnection()` if connection-creation fails) before executing the leak check.

The `SimpleConnectionState.borrow()` method needs to remove the unused `connectionCreateTimeout` argument. Since this method is only used internally by `SimpleURIConnectionPool`, it is not a breaking change.

Also, some refactorings and documentation updates can be done to clarify the code. These refactorings are at the SimplePool internal implementation level, and so are not breaking changes.

Finally, to coincide with re-addition of connection leak checking in the `restore()` method, and the change in its placement in both `borrow()` and `restore()`, the connection leak checking mock test has been updated (specifically, the `MockTimeoutLeakedConnection` class has been updated) to provide more information on connection leak finding and fixing.

### DIFF
--- a/client/src/main/java/com/networknt/client/simplepool/SimpleConnectionState.java
+++ b/client/src/main/java/com/networknt/client/simplepool/SimpleConnectionState.java
@@ -80,7 +80,9 @@ import java.util.concurrent.ConcurrentHashMap;
  *
  *   Users must provide a fixed 'now' value for the current time.
  *   This freezes a single time value for all time-dependent properties.
- *   This is important when calculating an aggregate state based on the values of 2 or more time-dependent states.
+ *   This is important in methods that calculate a state based on the results of multiple reads from one or more
+ *   time-dependent variables over the course of executing the method. For consistency of the calculation, its important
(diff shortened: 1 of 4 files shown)


AkashWorkGit (approved):

jaydeepparekh1311 (approved):

miklish:
hi @stevehu - fyi: merge conflict resolved

PR #2080, opened 2024-01-18
ExternalServiceHandler logging fix

KalevGonvick:
- Moved HandlerUtils to Handler module
- Added new method to populate audit logging fields wherever needed
- ExternalServiceHandler now populates the endpoint field in the audit log correctly.

### DIFF
--- a/handler/src/main/java/com/networknt/handler/HandlerUtils.java
+++ b/handler/src/main/java/com/networknt/handler/HandlerUtils.java
@@ -1,9 +1,13 @@
-package com.networknt.router.middleware;
+package com.networknt.handler;
 
+import com.networknt.httpstring.AttachmentConstants;
+import com.networknt.utility.Constants;
 import com.networknt.utility.StringUtils;
+import io.undertow.server.HttpServerExchange;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


handler/src/main/java/com/networknt/handler/HandlerUtils.java
```diff
@@ -68,4 +72,15 @@ public static String toInternalKey(String method, String path) {
         return String.format(INTERNAL_KEY_FORMAT, method, HandlerUtils.normalisePath(path));
     }
 
+    public static void populateAuditAttachmentField(final HttpServerExchange exchange, String fieldName, String fieldValue) {
+        Map<String, Object> auditInfo = exchange.getAttachment(AttachmentConstants.AUDIT_INFO);
+        if(auditInfo == null) {
+            auditInfo = new HashMap<>();
+            auditInfo.put(fieldName, fieldValue);
+        } else if (!auditInfo.containsKey(Constants.ENDPOINT_STRING)) {
```
stevehu, line 80: Should Constants.ENDPOINT_STRING be fieldName? Although we know this method is used to populate the endpoint, we must allow other users to use the same method to populate other fields.

Issue #2089, opened 2024-01-26
User Generic Name for SimplePool Connection-creation Threads (1.6.x)


miklish:
Currently SimplePool connection-creation threads are named either Callback-HTTP11 or Callback-HTTP11 depending on the type of connection created.

However, since this threadpool is only named ONCE, then the name of all SimplePool connection-creation threads is based on the first type of connection that SimplePool was asked to make (HTTP/1.1 or HTTP/2).

This can cause misleading in the logs, since if subsequent calls to the pool create different connection types, then the thread name anc connection type won't match.

E.g.: If an HTTP/1.1 connection is the first connection the pool makes, then any subsequent HTTP/2 connection-creation threads will have the name Callback-HTTP11.

stevehu:
My research into the git history shows this was changed from the conditional "Callback-HTTP2" / "Callback-HTTP11" naming in commit 820c20be back in March 2024. This ensures that even if the pool handles mixed traffic, the thread names remain accurate and generic. PR is merged.

PR #2090, opened 2024-01-26
Use Generic Name for SimplePool Connection Threads (1.6.x)

miklish:
Issue: https://github.com/networknt/light-4j/issues/2089

Currently SimplePool connection threads are named either "Callback-HTTP11" or "Callback-HTTP2" depending on the type of connection.

However, since this threadpool is only named ONCE, then the name of all SimplePool connection threads is based on the first type of connection that SimplePool was asked to make (HTTP/1.1 or HTTP/2).

This can cause misleading messages in the logs, since if subsequent calls to the pool create different connection types, then the thread name and connection type won't match.

E.g.: If an HTTP/1.1 connection is the first connection the pool makes, then any subsequent HTTP/2 connection threads will have the name "Callback-HTTP11".


AkashWorkGit (approved):

stevehu:
@miklish Are we ready to merge this PR? Thanks.

miklish:
@stevehu yes this can be merged thank you

PR #2110, opened 2024-02-07
fixes #2109 update basic, apikey and simple web token security handle…

### LINKED ISSUE #2109: update basic, apikey and simple web token security handler to be enabled by default
This will ensure that the security is turned on whenever the handler is in the request/response chain so that security is not bypassed accidentally. When UnifiedSecurityHandler is used, we check this flag to call the security handler if enabled.

stevehu:
…r to be enabled by default


DiogoFKT (commented):
Looks good, thanks!

PR #2171, opened 2024-03-07
Add validation to matchPathToPattern method

syntheshad:
Added a length validation to check if the base path and the request path are of equal length.

This is to avoid index out of bound issues that can be encountered for method rewrite rules.


stevehu (approved):

stevehu:
@syntheshad Thanks a lot for your help.

PR #2232, opened 2024-04-26
fixes #2231 Router rewriteMethod property does not work with path prefix

### LINKED ISSUE #2231: Router rewriteMethod property does not work with path prefix
`router.methodRewriteRules` property only works with full path because StringUtils must match string.split("/") item count between full path against configured path prefix

https://github.com/networknt/light-4j/blob/master/utility/src/main/java/com/networknt/utility/StringUtils.java#L1499-L1503

Called here:

https://github.com/networknt/light-4j/blob/2.1.33/proxy-handler/src/main/java/com/networknt/handler/ProxyHandler.java#L627

We need the property to allow only path prefix so to apply to any endpoint that matches the path prefix

stevehu:


DiogoFKT (commented):
Looks good! Thanks!

DiogoFKT (approved):

stevehu:
A test case that covers one of the use case. 

```
public class MethodRewriteRuleTest {

    @Test
    public void testMethodRewriteRule() {
        MethodRewriteRule methodRewriteRule = new MethodRewriteRule("/gateway/sit/service-request/{tracing-no}", "PUT", "PATCH");
        String requestPath = "/gateway/sit/service-request/123?version=1.1";
        Assert.assertTrue(StringUtils.matchPathToPattern(requestPath, methodRewriteRule.getRequestPath()));
        requestPath = "/gateway/sit/service-request?version=2";
        Assert.assertFalse(StringUtils.matchPathToPattern(requestPath, methodRewriteRule.getRequestPath()));
    }
}
````

PR #2288, opened 2024-07-09
fixes #2287 -Dlight-4j-config-password is not working for AutoAESSalt…

### LINKED ISSUE #2287: -Dlight-4j-config-password is not working for AutoAESSaltDecryptor
Currently, only the environment variable is supported for both lower and upper case. We must add System.getProperty() to overwrite the environment variable.

stevehu:
…Decryptor


DiogoFKT (commented):
Looks good! Thanks

PR #2340, opened 2024-09-18
fixes #2339 allow the req or res body encoding to be customized per p…

### LINKED ISSUE #2339: allow the req or res body encoding to be customized per path prefix
for transformer interceptors.

stevehu:
…ath prefix


DiogoFKT (commented):
Looks good! I'd add an extra unit test case for extra space at the end of pathPrefix line to test if required .isEmpty() check for `encoding` or not

PR #2346, opened 2024-09-19
fixes #2345 Fix the transformer matching with encoding

### LINKED ISSUE #2345: Fix the transformer matching with encoding
For path prefixes with encoding, we need to match only the first part with the request path.

stevehu:


response-transformer/src/test/java/com/networknt/restrans/ResponseTransformerInterceptorTest.java
```diff
@@ -144,4 +146,47 @@ public void run() {
...
     }
 
+
+    @Test
+    public void testMatch() {
+        List<String> list = new ArrayList<>();
+        list.add("/corp/mras/1.0.0");
+        list.add("/corp/lab/1.0");
+        list.add("/gateway/partyInfo/1.0 ISO-8859-1");
+        String requestPath = "/gateway/partyInfo/1.0/salesforce";
```
DiogoFKT, line 156: Can we add one more test case where `requestPath` is `/corp/lab/1.0` ? So we cover the scenario where there's no custom encoding config and pathPrefix is still found

stevehu: Done. Thanks.

PR #2355, opened 2024-10-08
2354 add expect 100 continue support

KalevGonvick:
Added Expect100ContinueHandler.
 - Sends back 100 continue when an Expect header contains 100-continue response.
 - Paths added to the ignore list will have their Expect headers removed, and the 100 continue header will be ignored.


stevehu (approved):
Looks good to me.

PR #2361, opened 2024-10-18
fixes #2360 invalid jwt token with no exp claim giving 500 response

### LINKED ISSUE #2360: invalid jwt token with no exp claim giving 500 response
The checkExpiry in JwtVerifier has the following line that can trigger NPE if claim.getExpirationTime returns null. 

```
if ((NumericDate.now().getValue() - allowedClockSkew) >= claim.getExpirationTime().getValue()) {
```

To fix it, add the NullPointerException in the cached exception.

stevehu:


DiogoFKT (commented):
Looks good thanks!

PR #2387, opened 2024-11-22
fixes #2386 add retry for external service handler

### LINKED ISSUE #2386: add retry for external service handler
To work around the Netskope timeout issue.

stevehu:


ingress-proxy/src/main/java/com/networknt/proxy/ExternalServiceHandler.java
```diff
@@ -201,12 +210,18 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
...
                         }
                     }
-                    return;
-                }
-            }
-        }
-        if(logger.isDebugEnabled()) logger.debug("ExternalServiceHandler.handleRequest ends.");
-        Handler.next(exchange, next);
+                })
+                .exceptionally(e -> {
```
DiogoFKT, line 214: Perhaps we want to confirm it's an IOException thrown here? According to JDK17 doc the completableFuture retruns with `exceptionally` when IOException OR SecurityException 
So I don't think we want to retry on SecurityException here

https://help.intrexx.com/apidocs/jdk17/api/java.net.http/java/net/http/HttpClient.html

PR #2454, opened 2025-02-08
Added metrics attachment for handler execution.

KalevGonvick:


KalevGonvick:
I need to test multiple requests at the same time. I think I may need to add some identifiers to the attachment to make it easier to tell which exchange the statement belongs to.

PR #2458, opened 2025-02-14
2452 create config file generator so configurations are always in sync with the pojo

stevehu:


stevehu:
found the following error on my local. 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project audit-config: Fatal error compiling: javax.lang.model.type.MirroredTypeException: Attempt to access Class object for TypeMirror java.lang.String -> [Help 1]
```

PR #2501, opened 2025-05-24
Change default value of enabled to false and fix a bug

atmoshaman:
## handler.yaml
Due to the code in the following screenshot, I believe the default value in the handler.yaml file of the handler-config module should be set to false; otherwise, it will affect the loading of the HandlerProvider interface implementation class in service.yml.

Server.java:
image

## NullPointerException

Avoid null pointer exceptions when the exclusionConfigFileList property is not configured in config.yml.


stevehu:
Thanks a lot for your help.

PR #2510, opened 2025-06-04
2509 header handler response not working as intended

KalevGonvick:
- Changed HeaderHandler logic to modify response headers after response is complete.
- Updated test cases to include edge case.
- Updated other test cases to include response conduit changes.


DiogoFKT (commented):
Shouldn't we be checking if  `pathPrefixConfig.getResponse();` exists before overriding `StreamSinkConduit` ? I believe this would reduce redundant processing for use cases where only request headers manipulation is required

KalevGonvick:
That is true, I will add this in a new PR.

PR #2514, opened 2025-06-10
2513 update token limit config with schema generation

KalevGonvick:
- Added ConfigSchema annotations to token-limit module.
- Updated Client Config to match original yml.
-  Refactored default value to be String type to simplify field annotation rules for generating yaml configs and json schemas.


stevehu (approved):

DiogoFKT (commented):
looks good to me

PR #2534, opened 2025-08-09
Find IP from system property STATUS_HOST_IP

atmoshaman:
Find IP from system property STATUS_HOST_IP.


stevehu (approved):

stevehu (commented):
The cache key should have both serviceName and envTag as one instance might be configured to route based on the envTag from the client to support multi-tenancy. What do you think?

atmoshaman:
> The cache key should have both serviceName and envTag as one instance might be configured to route based on the envTag from the client to support multi-tenancy. What do you think?

I have dropped the last commit regarding the modification of the cache key, as I believe that change was problematic. I will re-commit some code for that issue after this PR is merged.

stevehu:
@atmoshaman Thanks a lot for your help.

PR #2536, opened 2025-08-10
fixes fix cache key

atmoshaman:


atmoshaman:
> This ConsulRegistry is used by one of the customers for years already with 1.6.x branch, and it is working so far. The master branch is synced from the 1.6.x branch, and the logic is the same. Could you please let me know why do you think there is a bug? Thanks.

I have setup a service named `auth-center-service-1`.

My service in consul:

image

Client:

```
ServiceDef serviceDef = new ServiceDef("https", "com.kuibot.auth-center-service-1-1.0.0", "test", null);
        Http2ServiceRequest request = new Http2ServiceRequest(serviceDef, "/users", HttpVerb.GET);
        request.callForTypedObject(Map.class).thenAccept(user -> {
            System.out.println("111111");
        }).get();
```

The message of the stack error is as follows:

image

After tracing the code, I found that ConsulRegistry has some problems. The cache key is not correct in ConsulRegistry.java.

The key used to store the available service address in the `serviceUrls` map is the `serviceName`, without including the `environment` tag.

image

In the updateServiceCache method, the key used to retrieve data from the serviceUrls map is `serviceName + "|" + enviroment`, which will cause the serviceCache to never update.

image

To solve this problem, I can also not pass the enviroment parameter when initiating the call, which can avoid the problem mentioned above. However, this will introduce a new and more serious problem. If defines multiple environments for the same service, using only the serviceName to query in Consul will return multiple environments with the same name service, which will cause service confusion and this problem is very serious.

stevehu:
@atmoshaman Thanks a lot for your detailed explanation and it makes sense.

PR #2537, opened 2025-08-10
enhances correlation module

atmoshaman:


atmoshaman:
I saw a call to the MDC.clear() method within the ExceptionHandler, so I closed this PR.

PR #2542, opened 2025-08-27
[Draft] Negative Metrics Fix

KalevGonvick:
- Removed decrement logic
- Changed first time metric delta to be the count rather than -1
- Added overflow unit test

### DIFF
--- a/metrics-config/src/test/java/com/networknt/metrics/APMAgentReporterTest.java
+++ b/metrics-config/src/test/java/com/networknt/metrics/APMAgentReporterTest.java
@@ -4,9 +4,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.networknt.config.Config;
 import com.networknt.exception.ClientException;
 import com.networknt.handler.MiddlewareHandler;
-import io.dropwizard.metrics.MetricFilter;
-import io.dropwizard.metrics.MetricName;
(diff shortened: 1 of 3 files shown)


metrics-config/src/test/java/com/networknt/metrics/APMAgentReporterTest.java
```diff
@@ -337,30 +319,165 @@ public void overflowCounterTest() throws InterruptedException {
...
+    public void histogramMetricTest() {
+        MockTimeSeriesDbSender sender = new MockTimeSeriesDbSender("testService", "testProduct");
+        MetricRegistry registry = new MetricRegistry();
+
+        APMAgentReporter reporter = APMAgentReporter
+                .forRegistry(registry)
+                .convertRatesTo(TimeUnit.SECONDS)
+                .convertDurationsTo(TimeUnit.MILLISECONDS)
+                .filter(MetricFilter.ALL)
+                .skipIdleMetrics(false)
```
DiogoFKT: Shouldn't this be set to `true` ? When `false` asserting `sender.hasSeriesData()` is not enough since it could be getting idle data only and not the delta, right?

KalevGonvick: I will take another look at this

stevehu (approved):

PR #2713, opened 2026-03-25
fixes #2712 refactor the MultiThreadRuleExecutor to enable config reload

### LINKED ISSUE #2712: refactor the MultiThreadRuleExecutor to enable config reload

stevehu:


stevehu:
The changes by the other AI agent are a solid improvement. Here's my review:

✅ What's good — the 

RuntimeState
 pattern is excellent:

Immutable snapshot (

RuntimeState
 inner class) — Bundles endpointRules, rules, and ruleEngine into a single immutable object. This is a textbook approach for safe publication in concurrent code. When a config reload happens, all three fields are atomically swapped as one unit by assigning a new 

RuntimeState
 to the single volatile field. With our original approach, three separate volatile fields could theoretically be read in an inconsistent mix (e.g., old endpointRules with a new ruleEngine from a concurrent reload).


buildRuntimeState()
 is side-effect-free — It uses local variables (localRules, localEndpointRules, localRuleEngine) and returns a new 

RuntimeState
. This means the current state is never partially updated during construction — it's either the old state or the fully-built new state.

Correct ordering in 

checkConfigReload()
 (lines 78-80) — state = newState is assigned before config = newConfig. This ensures that any thread that sees the new config reference will also see the new state. Good attention to happens-before semantics.

currentState snapshot in 

executeRules(String, String, Map)
 (line 251) — The state volatile is read once into a local variable and then passed through the call chain. This guarantees the same 

RuleEngine
 and endpointRules are used for the entire execution of that request, even if a reload happens mid-flight.


loadPluginClass
 / 

loadActionClass
 now take explicit parameters — No more reliance on instance fields during construction. Clean and testable.

Test constructor (line 87) — Wraps the injected mocks into a 

RuntimeState
 with an empty endpointRules, so 

checkConfigReload()
 works correctly with the config == null guard.

⚠️ One minor observation (not a bug):

In 

executeRule()
 (line 153), 

getRuleEngine()
 reads the state volatile a second time after 

checkConfigReload()
. This is safe but slightly inconsistent with the 

executeRules
 overloads which snapshot state once. It would be marginally cleaner to do RuntimeState currentState = state; RuleEngine ruleEngine = currentState.ruleEngine; — but in practice this makes no functional difference since a second reload between those two lines is benign (you'd just use the newer engine, which is also valid).
Verdict: These changes are well-reasoned and correct. The 

RuntimeState
 pattern is the right approach for atomically swapping multiple related fields in a concurrent environment. I'd approve this as-is. Let me verify the tests pass:

PR #2723, opened 2026-03-27
fixes #2722 refactor the portal-registry for websocket controller

### LINKED ISSUE #2722: refactor the portal-registry for websocket controller

stevehu:


stevehu:
What changed:

  - discovery cache keys are now protocol-aware, so http and https entries for the same serviceId|tag no longer collide
  - subscription keys are also protocol-aware, so unsubscribe/replay tracks the correct subscription
  - microservice socket disconnects now trigger per-service re-registration instead of forcing a discovery-socket reconnect cycle
  - discovery URL rebuilding now preserves version and falls back to envTag when rebuilding parameters
  - PortalRegistryClient got backward-compatible default methods back, so the interface change no longer forces all external implementations to update immediately

  I also added regression coverage in light-4j/portal-registry/src/test/java/com/networknt/portal/registry/PortalRegistryTest.java and light-4j/portal-registry/src/test/java/com/networknt/portal/
  registry/PortalRegistryUtilsTest.java.

PR #2751, opened 2026-04-29
fixes #2750 light-gateway test has intermittent race behavior

### LINKED ISSUE #2750: light-gateway test has intermittent race behavior
The root cause is in light-4j, not the GatewayProxyTest assertion.

  The failing path is:

  light-gateway/src/test/java/com/networknt/proxy/GatewayProxyTest.java:186 borrows an HTTP/2 client connection because server HTTP/2 is enabled. The /post path also enables response
  body injection in light-gateway/src/test/resources/config/values.yml:30.

  Then light-4j/handler/src/main/java/com/networknt/handler/conduit/ModifiableContentSinkConduit.java:203 detects HTTP/2 and dispatches the actual response write to a worker thread:

  light-4j/handler/src/main/java/com/networknt/handler/conduit/ModifiableContentSinkConduit.java:213

  That worker thread calls next.write(...) directly against Undertow’s HTTP/2 framed stream. Undertow framed channels are sensitive to write/flush state and are effectively single-
  threaded around frame body state. Sometimes the worker writes while the HTTP/2 frame is not ready, so Undertow throws:

  UT000155: Framed channel body was set when it was not ready for flush

  After that, the response stream resets/times out, so the test receives no response body attachment. The NPE at light-gateway/src/test/java/com/networknt/proxy/GatewayProxyTest.java:211
  is only a symptom.

  The proper fix is in light-4j: rewrite the HTTP/2 branch of ModifiableContentSinkConduit so response writes are serialized through Undertow’s write readiness/IO-thread model instead of
  dispatching raw writes to a worker thread. The short-term CI fix is to disable server-side HTTP/2 for this gateway test config, since the backend proxy path is already HTTP/1.1 with
  proxy.http2Enabled: false.

  I ran the narrow test once with port permissions and it passed, which matches the intermittent race behavior.

stevehu:


stevehu:
I have reviewed the issue description and the PR diff.

The PR looks very solid and addresses the root cause perfectly. Here is a summary of my review:

Root Cause Analysis
The issue correctly identifies that Undertow's HTTP/2 FramedChannel is highly sensitive to concurrent or out-of-band write operations. Dispatching raw writes to an XnioWorker thread (as the old executeHttp2WriteThread did) violates Undertow's IO-thread model, leading to race conditions where the frame body is set when the channel isn't ready for a flush, throwing UT000155.

PR Implementation Review
The PR correctly rewrites the http2Write path to use an asynchronous write loop tied to Undertow's native write readiness notifications.

Here are the key aspects that are handled correctly:

Thread Safety & IO-Thread Alignment: The introduction of runOnWriteThread ensures that all state mutations (like advancing buffer indexes and setting boolean flags) and conduit write operations happen strictly on the channel's associated IO thread. This completely eliminates the race condition.
Proper WriteReadyHandler implementation: The handler is registered correctly, and drainHttp2Response() properly handles partial writes. If next.write(buffer) returns 0, it correctly backs off, calls next.resumeWrites(), and waits for the next IO thread notification.
Correct Flush & Termination Lifecycle: The completeHttp2Response() method correctly manages the termination of writes. It calls next.terminateWrites(), then loops on next.flush(). If the flush isn't ready, it resumes writes to wait for the next event, which perfectly aligns with Undertow's conduit flush expectations.
Memory Management / Leak Prevention: closeRemainingBuffers() safely cleans up any unwritten pooled buffers if the write loop fails, completes, or is forcibly terminated, ensuring no memory leaks occur in the PooledByteBuffer instances.
Robust Error Handling: The failHttp2Response properly catches any IO exceptions during the write process, ensures resources are released, and calls next.truncateWrites() to forcefully close the stream if things go wrong.
Conclusion
This is a textbook implementation of an asynchronous Undertow conduit write loop. The changes are logically sound, clean, and safe to merge. Great work!

Issue #257, opened 2018-08-13
flatten the config files into the same directory for k8s


stevehu:
By default, we have most config files in the config folder but some files are in tls and oauth folder as these are the referred files. When creating k8s Secrets, kubectl can support only one directory but not sub-directory. There is an issue opened for a long time to support a hierarchical structure but haven't resolved yet. Meanwhile, it makes sense for us to flatten the config file structure to move all files to the config folder. 

https://github.com/kubernetes/kubernetes/issues/62421

PR #2780, opened 2026-09-09
Fix mixed map and POJO configuration cache reads

### LINKED ISSUE #2779: Config cache returns incompatible types when map and POJO loads are mixed
Mixed map and typed configuration loads return the wrong type or throw `ClassCastException`, depending on which representation first populated the cache.

Reported while testing networknt/light-kafka#258, specifically https://github.com/networknt/light-kafka/pull/258#issuecomment-5606252102.

Reproduction with a configuration that can be deserialized as `MyConfig`:

```java
Config config = Config.getInstance();
config.clearConfigCache("example");
config.getJsonMapConfig("example");
Object uncached = config.getJsonObjectConfigNoCache("example", MyConfig.class); // MyConfig
Object cached = config.getJsonObjectConfig("example", MyConfig.class); // Map, not MyConfig
```

Reversing the order (cache a POJO first, then call `getJsonMapConfig`) throws `ClassCastException`. The default-loader APIs have the same problem.

The cache is keyed only by configuration name. Cached object reads do not validate the requested class, and cached map reads cast the entry directly. Uncached object reads deserialize into the requested class. This explains the observed type difference without requiring any change to the configuration file.

Expected: mixed map/POJO reads should return the requested representation in either order, while compatible cache hits and explicit cache invalidation continue to work.

Reproduced with framework regression tests and a Kafka regression against PR #258 at `def6fd56144d216d6af1c9f8b7bfa03067aafafc`.

stevehu:
Fixes #2779. Related: https://github.com/networknt/light-kafka/pull/258.

Mixed map and POJO configuration reads must retain stable identities until invalidation. The original name-only cache returned incompatible objects; the initial PR workaround avoided that collision by repeatedly reloading the second representation, breaking identity-based singleton reload detection.

Replace that workaround with per-name cache entries containing representations keyed by actual Java `Class` identity. `Map.class` object reads share the map entry; with a custom loader, the first reader selects loadObjectConfig or loadMapConfig according to the API it called. Configured/default accessors retain first-load-wins sharing. Cache hits avoid loading locks; top-level misses coordinate per name, without holding a global lock across I/O. Re-entrant load callbacks reuse completed values but resolve cold nested reads in detached entries, avoiding acquisition of a second shared entry lock. Nested misses can repeat loader work and do not publish into the normal cache; top-level single-load and identity guarantees remain intact.

Clearing, replacing, or switching class loaders detaches old entries. Loads already in flight may finish for their existing callers but cannot repopulate the active cache. Class-loader switches discard loaded/derived representations while preserving explicit putInConfigCache injections, which may have no reloadable source. References owned by those explicit injections remain until replacement or explicit invalidation.

For the default non-exclusion path without a configured custom loader, derive POJOs from the cached merged map using the existing Jackson conversion semantics. Exclusions preserve SnakeYAML direct binding; custom loaders retain their object binding. Cache all successful representations, reject incompatible loader output before insertion, and lock both YAML object parsing branches, including uncached reads.

Loader initialization happens before configured cache misses take entry locks. Constructor-time configuration reads use detached bootstrap entries and the default loader, preserving existing cached/injected settings without recursively constructing the loader or publishing default data into its normal cache. Cache hits do not force loader initialization. Derived binding errors retain their cause and now log and identify the configuration name.

Missing/failed loads release empty entries only after all waiting readers finish. This prevents retention of unresolved names while preserving single-entry identity for concurrent retries.

Semantics retained explicitly:
- Mutable maps stay shared. POJOs capture values at conversion time; later map mutations do not update existing POJOs.
- Custom-loader and exclusion representations can load independently; this does not promise an atomic cross-representation source snapshot for those paths.
- Public NoCache methods continue to bypass the cache.
- Kafka's fallback for older framework versions remains in place.

Regression coverage includes mixed load order and identity, edited/invalid source files, map mutation and replacement, exact-name invalidation, class-loader isolation, configured/default sharing, custom-loader validation, concurrent initialization and invalidation, and uncached YAML parsing. Cross-module tests cover HandlerConfig singleton identity before/after invalidation and preservation of app-status entries through typed reads.

Validation:
- A depth-guarded two-thread mutual cross-read regression times out on `d144fa0e1` and passes with the re-entrant guard, covering both map and object callbacks and preservation of final root identities.
- All five follow-up review regressions fail on `3868cc55d` (including a timeout in the forced loader-construction deadlock) and pass with the fixes.
- Three deterministic regressions fail against the original PR commit `a186402b1` (identity, invalid-source reread, incompatible loader result) and pass with this replacement.
- Config suite: 97 run, 0 failures, 5 skipped. Handler-config suite: 8 passed.
- All 20 cache regressions and the singleton regression passed. The preceding revision also passed the clean security suite (72 run / 3 skipped) and server suite (16 run / 2 skipped).
- Full light-kafka reactor at `60032e8`, using the locally installed patched config dependency: 65 passed.
- Pre-commit and diff checks passed. GitNexus follow-up analysis reports HIGH impact across 9 shared configuration flows.

An initial broader run encountered five security-test failures from generated client/security fixtures overwritten by `JwtVerifierObservabilityTest`. The same failures reproduce on `a186402b1` on a repeated run; the clean security suite passes with the replacement. This unrelated fixture-isolation issue is outside the PR.

Release ordering: merge and publish the patched light-4j config dependency before qualifying/merging the dependency-sensitive Kafka regression. The local installation proves compatibility but is not verification of a published snapshot; that check remains pending publication.


stevehu:
Code review(high · 6 findings)
light-4j/config/src/main/java/com/networknt/config/Config.java
  ● 254 [race-condition] The incompatible-type fallback calls the object loader outside `synchronized (FileConfigImpl.class)`, and `loadSpecificConfigFileAsObject` (lines 473/476) drives the process-wide, non-thread-safe SnakeYAML `Yaml` instance with no lock — unlike `loadSpecificConfigFileAsMap:508`, which does `synchronized (getYaml())`. Because the fallback now fires on every call rather than once at startup, concurrent config reads corrupt each other.
  ● 253 [correctness]    When the name-only cache holds an incompatible type, the requested type is re-read from disk on every single call and never cached. Two consequences: object identity is lost, and the two representations of the same file drift apart, because the cached one is frozen at first load while the other tracks the current file contents — silently bypassing light-4j's explicit `clearConfigCache`/reload coordination.
  ● 281 [error-handling] Re-reading on every call re-exposes the caller to load failures. `loadSpecificConfigFileAsObject:481` converts any parse/IO error into `throw new RuntimeException`, so a config file that becomes unreadable or invalid after startup now propagates an exception out of a call that previously returned the long-cached value.
  ● 342 [correctness]    In the mixed case `getJsonMapConfig` returns a fresh map per call, so in-place mutations by callers are no longer visible to other readers. light-4j's own idiom relies on this sharing: `server-config/src/main/java/com/networknt/server/ModuleRegistry.java:100` (`registerPlugin`) calls `maskNode(config, mask)`, which mutates the caller-supplied map in place, and callers routinely pass `Config.getInstance().getJsonMapConfig(CONFIG_NAME)`.
  ● 247 [correctness]    The loaded value is cached at line 247 before the new `clazz.isInstance` check at 253 runs. If the loader returns something that is not an instance of `clazz`, that value is stored in the cache and the very same call immediately reloads — and every later call reloads too, permanently. Guarding the `configCache.put` with the same `clazz.isInstance` test avoids poisoning the cache.
kafka-common/src/test/java/com/networknt/kafka/common/KafkaConfigApiCompatibilityTest.java
  ●  33 [build-break]    PR #258 adds a regression test whose fix lives entirely in an uncommitted light-4j change. light-kafka's `version.light-4j` is the moving `2.3.8-SNAPSHOT`, so until the Config.java change is committed, released, and the snapshot republished, any CI or clean local build of light-kafka resolving the published snapshot fails.

Answer to your two questions

Does the light-4j change make sense given the PR comments? The diagnosis is right, and I confirmed it empirically. KafkaStreamsConfig.load() populates configCache under the bare name "kafka-streams" with a LinkedHashMap; a later getJsonObjectConfig("kafka-streams", KafkaStreamsConfig.class) then blind-casts that entry. On the unpatched code my 16-thread probe hit ClassCastException: LinkedHashMap cannot be cast to TestConfig 4800/4800 times — exactly what KalevGonvick observed, and it explains why getJsonObjectConfigNoCache disagreed with getJsonObjectConfig. So a light-4j fix is genuinely needed; light-kafka can't fix this in its own repo (its KafkaConfigUtils.getJsonMapConfig:60 CCE-catch is already a workaround for the same defect).

Does it introduce risks? Yes — the specific shape of the fix is the problem, not the goal. Rather than making the cache type-aware, it leaves the cache name-only and makes the losing representation permanently uncached. That converts a loud, deterministic failure into four quiet ones, all verified by running probes against both HEAD and the stashed baseline:

- light-4j/config/.../Config.java:254 — the fallback reload runs outside synchronized (FileConfigImpl.class) and parses on the shared, non-thread-safe SnakeYAML Yaml with no lock (unlike the map loader at :508). 16 threads × 300 calls → 38/4800 corrupted parses (StreamStartToken ... null, unconstructable recursive node, Push back buffer is full, field "alue"). Wrapping lines 472–478 in synchronized (getYaml()) → 4800/4800 clean.
- light-4j/config/.../Config.java:253 — the losing type is re-read from disk on every call and never cached: identity is lost, and after an on-disk edit with no reload the cached map still returns v1 while getJsonObjectConfig returns v2. Two modules in one JVM disagree about the same file, bypassing the explicit reload path.
- light-kafka/kafka-common/src/test/.../KafkaConfigApiCompatibilityTest.java:33 — with the light-4j patch stashed and reinstalled as the published 2.3.8-SNAPSHOT, this test is a BUILD FAILURE. The light-4j commit must land and publish before #258 merges.
- light-4j/config/.../Config.java:281 — a config file edited into an invalid state after startup now throws RuntimeException out of a live call path that previously served the cache.
- light-4j/config/.../Config.java:342 — in-place mutations of the map from getJsonMapConfig are dropped; ModuleRegistry.registerPlugin:100's maskNode relies on that sharing, so masking now sticks or not depending on cache order.
- light-4j/config/.../Config.java:247 — the load result is cached before the isInstance check, so a ConfigLoader SPI returning a non-clazz value poisons the cache and reloads forever.

Suggested direction: key the secondary representation by configName + '#' + clazz.getName() (or #map) instead of not caching it, and make clearConfigCache(name) sweep that prefix. That preserves caching, identity, single-load-per-reload semantics and the existing lock discipline, and it removes findings 1, 2, 4, 5 and 6 together. Independently, loadSpecificConfigFileAsObject should take synchronized (getYaml()) regardless — getJsonObjectConfigNoCache already has that hole today.

Environment is back as I found it: probe tests deleted, patched config module reinstalled to ~/.m2, both working trees unchanged (Config.java modified + untracked ConfigMixedRepresentationTest.java in light-4j; modified KafkaConfigApiCompatibilityTest.java in light-kafka). The full light-4j config suite passes with the patch: 79 run, 0 failures, 5 skipped.

stevehu:
First, a note: the fix landed as a186402b1 (#2779) while I was working, so everything below is measured against that commit and its parent 8e836a53f.

1. Is nocache only used by the module registry? No.

106 *NoCache( call sites across light-4j, light-kafka, light-portal, light-rest-4j, light-websocket-4j, light-aws-lambda and oauth-kafka (main sources only):

- 82 are Config.getNoneDecryptedInstance().get*NoCache(...) feeding registerModule/registerPlugin — your design, working as intended.
- 8 are legitimate non-registry uses: values.yml bootstrap (ConfigInjection:81, DefaultConfigLoader:276, ConfigReloadHandler:71 — these must bypass the cache), reload() methods (KafkaProducerConfig:192, KafkaStreamsConfig:176), one commented-out line, and KafkaConfigUtils:63 (the CCE workaround for this very bug).
- ~16 are violations — primary config-load paths calling nocache:
  - light-4j/router-config: TokenConfig:57, PathPrefixServiceConfig:70, OAuthServerConfig:82, PathServiceConfig:69, RouterConfig:390, ServiceDictConfig:71 — all private XConfig(String configName) constructors doing mappedConfig = config.getJsonMapConfigNoCache(configName)
  - light-4j/restrans-config: ResponseFilterConfig:64, ResponseCacheConfig:64
  - light-4j/jaeger-tracing: JaegerConfig:48
  - light-4j: Mask:53, Server.mergeStatusConfig:531
  - light-websocket-4j: all three configs
  - light-rest-4j/openapi-meta: OpenApiHandler:120,132
  - light-aws-lambda: LambdaSchemaValidator:45

2. If everyone used only the cached version, would the issue remain? Yes — unchanged.

Two independent confirmations:

Structurally, the nocache methods can't participate. getJsonObjectConfigNoCache:304, getJsonMapConfigNoCache:400 and getDefaultJsonMapConfigNoCache:411 each return load*(...) directly — they never read from and never write to configCache. (getNoneDecryptedInstance() also has its own separate configCache instance field, so registry loads are doubly isolated.)

Empirically, on parent 8e836a53f, a probe touching only the cached accessors — zero NoCache calls anywhere — throws in both orders:

CACHEONLY mapThenObject=THREW ClassCastException
CACHEONLY objectThenMap=THREW ClassCastException
and on a186402b1 both return OK default config.

The real light-kafka reproduction already follows your discipline: KafkaStreamsConfig.load() uses the cached Config.getInstance().getJsonMapConfig(configName) (via KafkaConfigUtils:59) for the real load and reserves getNoneDecryptedInstance().getJsonMapConfigNoCache(...) for ModuleRegistry (KafkaStreamsConfig:162). The application then calls the cached getJsonObjectConfig("kafka-streams", KafkaStreamsConfig.class). Two cached accessors, one name-only key, two representations. KalevGonvick's getJsonObjectConfigNoCache observation was the diagnostic that exposed it, not the cause.

The actual root cause is that the typed-config refactor introduced a second cached representation of each file while configCache stayed keyed by name alone — a holdover from when everything was a HashMap.

A counterintuitive consequence worth weighing: enforcing your stated discipline more strictly would create more collisions, not fewer. jaeger-tracing is collision-free today only by accident — JaegerStartupHookProvider:28 caches a JaegerConfig POJO under "jaeger-tracing", and nothing caches a map there only because JaegerConfig:48 violates the rule by using nocache. Migrate that constructor to the cached accessor and jaeger-tracing becomes a second kafka-streams. Same for the six router-config classes. So the invariant that actually needs restoring is one config name → one cached representation, or a cache key that includes the representation (configName + '#' + clazz.getName(), with clearConfigCache(name) sweeping the prefix).

3. How to guarantee only the registry calls nocache

- Remove the need rather than police it. Add ModuleRegistry.registerModule(String configName, String moduleClass, List<String> masks) that performs the masked nocache load internally. That retires all 82 correct call sites' need to name a NoCache method, after which the remaining users are, by construction, the violations. Then @Deprecated(forRemoval=true) the public NoCache methods.
- Enforce in CI. An ArchUnit test in light-4j is the cheapest real gate; for cross-repo coverage, a light-bot grep gate failing on NoCache( outside an allowlist works too. Either way you need a seeded allowlist for the ~16 current violations plus permanent exemptions for reload() and the CENTRALIZED_MANAGEMENT bootstrap, then shrink it.
- True guarantee costs a breaking change. Java access control is the only hard enforcement: move the NoCache methods off the public Config API onto a separate type that ModuleRegistry obtains. Config is consumed by every downstream repo, so that's 2.4.0-scale.
- A StackWalker WARN in the three NoCache methods when the caller isn't ModuleRegistry/ConfigInjection/a reload method is useful for finding stragglers in live deployments, but it is not enforcement.

I'd do the first two. But none of them address the bug in this thread — that one lives entirely in the cached path, and my six earlier findings against a186402b1 (the off-lock shared-Yaml race being the most serious) still stand.

I've saved the design intent and the violation inventory to memory so future config reviews start from it.

stevehu:
I have to correct myself: my "~13 violations" bucketing under-credited your design badly. 50 config classes follow exactly the discipline you described — and they do something I'd missed.

The idiom I missed

HandlerConfig.load() (and 49 others — ClientConfig, AuditConfig, BodyConfig, AccessControlConfig, ConsulConfig, …):

Map<String,Object> mappedConfig = Config.getInstance().getJsonMapConfig(configName);   // CACHED
if (instance != null && instance.getMappedConfig() == mappedConfig) return instance;   // identity
synchronized (HandlerConfig.class) { ... instance = new HandlerConfig(configName);
    ModuleRegistry.registerModule(configName, ..., Config.getNoneDecryptedInstance().getJsonMapConfigNoCache(configName), null); }

Cached map for the real load, nocache map for the registry — precisely your intent. And the singleton is invalidated by reference identity of the cached map. That's how /adm/config/reload actually works: ConfigReloadHandler.reloadModule:97-99 does nothing but clearConfigCache(configName); the next getJsonMapConfig returns a different object, == fails, the singleton rebuilds and re-registers. (I'd started to conclude reload was a no-op — wrong. RELOAD_METHOD = "reload" at ConfigReloadHandler:41 is declared and never used, which is what put me on that path.)

What the true issue is

Not the cache key. Config.configCache is doing two jobs: a performance cache, and the invalidation channel for 50 config singletons via map reference identity. A name-only key was fine while there was exactly one representation per name. The typed-config refactor added a second representation; a186402b1 fixed the type collision by returning an uncached fresh object — which silently breaks identity stability, i.e. job two.

This makes my finding #5 the most serious one in the review, not a medium. For any of those 50 names that anything also reads via getJsonObjectConfig, the identity check is never true again, so load() re-reads the file, rebuilds the singleton, and re-runs ModuleRegistry.registerModule (with its deepCopy) on every call, inside a class-level lock — for classes like ClientConfig/AuditConfig/BodyConfig that sit in the request path. Silently. light-kafka is immune only by accident: KafkaConfigUtils.sameMappedConfig uses left == right || left.equals(right); light-4j uses raw ==.

So the ClassCastException was the symptom that shouted. This one doesn't.

Can the nocache loader return a Config class instead of a Map?

For the registry path, I'd say no — four reasons, the last being the one that matters most:

1. Lossiness. A typed POJO is a schema-limited projection; undeclared keys vanish. That is literally light-kafka #257. /adm/server/info exists to show what's actually in effect — including the keys being ignored. Making it lossy removes the thing you consult it for.
2. Masking would round-trip back to a map (convertValue(pojo, Map.class)), just a lossy one.
3. Undecrypted rendering. The registry uses getNoneDecryptedInstance() so CRYPT values show as ciphertext; a POJO path needs parallel machinery.
4. The map is not just transport — it's the invalidation token. getMappedConfig() == mappedConfig compares map identity. Switching representations isn't a local change to the registry; it touches the invalidation protocol of 50 classes.

The good idea to extract: annotation-driven masking. You already have @ConfigSchema/@StringField/@ObjectField. Add masked = true and ModuleRegistry can drop its List<String> masks parameter entirely — masks stop drifting from the config classes, and registerPlugin's in-place maskNode mutation (registerModule deep-copies at :64, registerPlugin at :100 does not) disappears with it.

Design recommendations, in priority order

1. Key the cache by name + representation, preserving identity. configName for the map, configName + '#' + clazz.getName() for each POJO; clearConfigCache(name) sweeps the prefix so one reload invalidates all representations coherently. This is the minimal change that fixes the real problem: it restores identity stability for the 50 classes, eliminates the per-call reload, closes the off-lock Yaml race, and keeps /adm/config/reload working.
2. Then replace == with an explicit generation counter. if (instance != null && instance.generation == Config.getInstance().getGeneration(configName)). Same O(1) check, but invalidation stops depending on an allocation detail — which is what made this failure invisible. Worth doing even with #1.
3. Annotation-driven masking, as above.
4. Give ModuleRegistry a name-based overload that does the masked nocache load itself, so 82 call sites stop hand-rolling it. Keep the map.
5. Converge the ~13 legacy classes (router-config ×6, restrans ×2, jaeger, websocket ×3, LambdaSchemaValidator) onto the 50-class idiom — today TokenConfig.load() re-reads the file every call and is unreachable by /adm/config/reload. Ordering trap: do this after #1, or you create new collisions.
6. Delete RELOAD_METHOD at ConfigReloadHandler:41.

The one thing I'd not do is push further on "nocache is registry-only" as the organizing rule. The evidence says the real invariant is narrower and more important: one config name must have one stable cached identity per representation. Get that right and the nocache question mostly stops mattering.

stevehu:
Verified the parts that decide this. Working trees are clean — probes removed.

Short answer

It would resolve the collision, and it would clean up more than you'd expect. But as the code stands there's one hard blocker, and it doesn't remove the need for the cache fix.

What it fixes — more than just the collision

With one class per name and every read going through getJsonObjectConfig(name, XConfig.class), clazz.isInstance always passes, the entry stays cached, identity stays stable. And load() collapses to:

public static XConfig load() { return (XConfig) Config.getInstance().getJsonObjectConfig(CONFIG_NAME, XConfig.class); }

No instance field, no synchronized, no getMappedConfig() == mappedConfig. One cache, one invalidation signal. That deletes the entire fragile identity-invalidation mechanism across 50 classes — genuinely better than what you have.

The hard blocker: the POJO path is strict-schema

CentralizedManagement.convertMapToObj:106-109 builds a bare new ObjectMapper() (FAIL_ON_UNKNOWN_PROPERTIES defaults to true), and 0 of the 91 config classes carry @JsonIgnoreProperties(ignoreUnknown = true). Same file, two accessors:

STRICT viaMap=OK hello
STRICT viaPojo=THREW UnrecognizedPropertyException: Unrecognized field "unknownKey" ... not marked as ignorable

loadSpecificConfigFileAsObject:481 wraps that into a RuntimeException — a startup failure. Today's map path is deliberately lenient: setConfigData() reads the keys it knows and ignores the rest. Converting all 91 turns every unknown key — a typo, a deprecated key still sitting in a file, a config-server injection, a forward-compat key from a newer light-4j — into a hard boot failure for every downstream application. That is light-kafka #257 generalized to the whole ecosystem.

It's a one-line fix (disable that flag, or @JsonIgnoreProperties on a shared base), but it has to be a deliberate decision made first. Worth flagging the irony: you introduced typed classes to get more validation, and this makes them reject what the map path tolerated — a blunter kind of validation than you were reaching for.

The map path can't actually go away

Some configs can't be POJOs: status.yml (open map of arbitrary codes, and Server.mergeStatusConfig mutates the cached map via putAll), OpenAPI specs, values.yml, config.yml itself, and the kafka properties map behind #257. So you'd shrink the collision surface a lot but not to zero — and mergeStatusConfig stays exactly as exposed as it is today. Related: light-portal already calls getJsonObjectConfig(name, Map.class) in six places (EmbeddingTaskStartupHook:49, GenAIPersistenceImpl:7549, OperationalCleanupStartupHook:94, and three config classes) — the object API returning a Map. Harmless now, since none of those names has a typed class, but it's the pattern you'd have to police forever.

The migration isn't mechanical

setConfigData() does coercion Jackson won't reproduce: Config.loadBooleanValue:789-800 accepts a Boolean or a String and maps empty-string to false; light-kafka needed explicit String.valueOf for a numeric group.id. Each of the 91 classes needs its coercion re-established or configs silently change meaning — and every downstream caller of getMappedConfig() breaks.

Recommendation

Do both, cache first.

- Cache fix now. Name+representation key with a prefix-sweeping clearConfigCache is ~20 lines and fixes the ClassCastException, the identity-invalidation regression, the off-lock Yaml race, and the per-call reload. Release-safe.
- All-POJO convergence as a multi-release program, gated on the strictness decision, done per-module, keeping the map path for open-ended configs.

The asymmetry that decides the order: the cache fix makes the system correct regardless of caller discipline; the migration makes it correct only as long as everyone stays disciplined. An all-POJO world with a name-only cache is still one getJsonObjectConfig(name, Map.class) call away from the same bug — and this time it fails silently rather than with a ClassCastException.

stevehu:
I reviewed all four substantive comments against PR #2780’s current head, a186402b1. The central criticism is correct: my fallback fixes the type collision but does not preserve caching
  semantics. No code changed.

  Points I accept:

  - Stable identity matters. HandlerConfig and other singletons compare cached map references to detect reloads. Returning a fresh map repeatedly can rebuild those singletons on every call.
  - Repeated loading causes drift and new failure exposure. Once loaded, both representations should remain cached until invalidation.
  - The YAML locking gap is real. Object parsing lacks the lock used by map parsing. Fixing the cache alone does not protect concurrent NoCache calls.
  - Loader results need validation before caching. A custom loader returning an incompatible object should produce a clear error without poisoning the cache or triggering repeated loads.
  - Kafka dependency ordering now matters. PR #258 has advanced to 60032e8 and now includes the mixed-loading regression. It needs the patched framework dependency available to CI.

  Some qualifications:

  - NoCache did not cause the collision; two cached accessors suffice.
  - registerPlugin mutates its supplied map, but preserving shared masking side effects is not the justification for this fix. Stable configuration identity is.
  - Separate representation caches prevent repeated drift after loading; they do not automatically guarantee identical source snapshots if the file changes between the first map and first
    POJO load.

  - I would avoid concatenated name#class keys and prefix deletion: structured keys avoid name collisions and distinguish classes from different class loaders.

  My proposed plan:

  1. Cache each representation with stable identity. Preserve the existing map cache behavior and add typed entries keyed by configuration name and actual Class. Ensure
     getJsonObjectConfig(name, Map.class) shares the map representation.

  2. Coordinate invalidation and replacement. Make clearConfigCache(name), clear(), and putInConfigCache(...) handle associated typed entries consistently. Synchronize cache population and
     invalidation so an in-progress load cannot republish an invalidated entry.

  3. Protect YAML parsing. Use the same YAML-instance lock for object and map parsing, including uncached paths. Keep merging outside that lock.
  4. Validate custom-loader output before insertion. Preserve null/missing-config behavior; reject incompatible non-null results explicitly.
  5. Expand regression coverage. Test both load orders, stable identity for both representations, map mutation visibility, invalidation/replacement, file changes after caching, custom
     loaders, concurrent reads, and a real singleton’s reload behavior.

  6. Validate and update the related PRs. Run affected framework tests and the Kafka reactor, then verify Kafka against the published patched dependency rather than relying only on a local
     installation.

  I would defer generation counters, all-POJO migration, annotation-driven masking, registry overloads, and broad NoCache cleanup to separate work. They are larger compatibility changes and
  are unnecessary to correct this cache defect.

  I’ll wait for your approval of this plan before editing.

stevehu:
Working trees are clean; no leftover probes. Here's my review.

Coverage check — all six findings are addressed

┌──────────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────┐
│                             Finding                              │                                   Plan item                                   │
├──────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ #1 off-lock shared Yaml race (Config.java:254, parse at 473/476) │ 3 ✅                                                                          │
├──────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ #2 losing representation never cached → identity loss + drift    │ 1, 2 ✅                                                                       │
├──────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ #3 light-kafka test fails against published snapshot             │ 6 ✅                                                                          │
├──────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ #4 unparsable-after-startup file now throws                      │ 1 (implicitly — no repeat reads), tested via 5 ✅                             │
├──────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ #5 map mutation visibility                                       │ 1 ("preserve existing map cache behavior") + 5 ✅                             │
├──────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ #6 non-clazz loader result cached then reloaded forever          │ 4 ✅ (and you go further — explicit rejection beats my "just don't cache it") │
└──────────────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────┘

Your three qualifications are all correct, including the two that push back on me. NoCache genuinely didn't cause the collision. And you're right that registerPlugin's masking side effect is a consequence to preserve, not the justification — stable identity is the real reason, and I'd rank it higher than I did initially.

Structured keys over my name#class + prefix sweep: your call is better. Class-loader identity is the argument I didn't make. One caveat: a strong Class reference in the cache pins its ClassLoader. Config.setClassLoader has no production callers in these repos today, so it's latent rather than live — but if a module loader ever swaps loaders, those entries leak. Either clear on swap or hold the Class weakly.

Gaps worth folding in

1. Derive the POJO from the cached map instead of re-reading the file. This dissolves your own caveat about snapshot divergence. For a non-exclusion config with no custom ConfigLoader, the two paths are provably identical: the map path is yaml.load + mergeMap(decrypt, m), and the object path is yaml.load + merge(decrypt, m) + convertValue(m, clazz). So the cached map is the merged map, and the POJO is exactly convertValue(cachedMap, clazz). Deriving gives you a guaranteed-identical snapshot, one file read instead of two, and zero additional YAML-parse exposure.

Preconditions to check at runtime — fall back to a fresh load when either fails:
- configLoader == null (a custom loader's loadObjectConfig isn't derivable from its loadMapConfig)
- !ConfigInjection.isExclusionConfigFile(configName) — the exclusion branch uses yaml.loadAs(inStream, clazz), SnakeYAML direct binding, whose coercion differs from Jackson's convertValue

2. Decide getJsonMapConfig vs getDefaultJsonMapConfig deliberately. Today they share one name-keyed entry, so whichever runs first wins and the other silently gets the wrong loader's result. Same for the two object variants. Preserve that parity for typed entries (I'd keep it — changing it is a separate behavior change), but make it a decision rather than a side effect of the new key shape.

3. Item 3 must cover the exclusion branch too. getYaml().loadAs(inStream, clazz) at :473 is a parse and needs the lock as much as getYaml().load(inStream) at :476. Keeping the merge outside is right and matches how loadSpecificConfigFileAsMap:508 already does it.

4. Item 2's synchronization — don't widen the lock to cover the load. The obvious implementation (hold the lock across load+put) reintroduces global serialization: I measured 4800 concurrent object loads at 0.96s unsynchronized vs 5.06s with the parse serialized. Prefer capturing a cache epoch before the load and doing a compare-and-put after, so loads stay concurrent while a concurrent clearConfigCache still wins. That's an internal detail, not the per-config generation counter you deferred.

5. Call out a win you're getting for free. Server.mergeStatusConfig:531 loads status.yml cached and mutates it via putAll so app-status codes stick. Today that silently breaks the moment anyone adds a typed StatusConfig. Under separate representation caches it stays correct permanently — worth an explicit test, since it's the highest-consequence instance of finding #5 (custom status codes vanishing with no log).

6. Two small ones. ConfigMixedRepresentationTest.freshConfig() reflects on Config.getInstance().getClass().getDeclaredConstructor(), which breaks under a ServiceLoader-provided Config — worth hardening while you're expanding item 5. And on the Kafka side, KafkaConfigUtils.getJsonMapConfig:60's ClassCastException catch becomes dead once the framework fix ships; leave it for version-straddling or remove it deliberately, but note it.

On the deferrals

Agreed on all of them, with one caveat: since 50 classes depend on getMappedConfig() == mappedConfig, item 5's "real singleton's reload behavior" test is the only thing standing between you and a silent regression there. Make that test assert both directions — that load() returns the same instance across calls with no invalidation, and a different one after clearConfigCache. That's the guard that lets you safely defer the generation counter.

The plan is sound and correctly scoped. With items 1–4 above folded in, I'd say go ahead.

stevehu:
Most of these additions make sense. I’ll incorporate them, with two refinements.

  - Derive POJOs from the cached map where equivalent. The code confirms the default, non-exclusion paths use the same merge logic. Reuse the existing Jackson conversion semantics without
    merging twice. For exclusions and custom loaders, retain their object-loading behavior and cache its result. Because loader initialization is lazy, checking only configLoader == null is
    insufficient; eligibility must consider the configured loader.

  - Preserve configured/default cache sharing deliberately. Changing their existing first-load-wins behavior belongs in separate work.
  - Lock both YAML parsing branches: load and loadAs. Keep conversion and merging outside the parsing lock.
  - Avoid adding a global load lock. Use coordinated loading per configuration and an invalidation token to prevent stale publication. An epoch check alone is insufficient: concurrent
    initial loads must also converge on one cached identity. The shared YAML parser still requires serialization regardless of the timing comparison.

  - Clear caches when the class loader changes. This releases cached class references and prevents outstanding loads from repopulating entries from the previous loader.
  - Strengthen tests. Include status-map additions surviving typed reads, both singleton identity assertions, invalidation during loading, exclusions, custom loaders, and configured/default
    sharing. Construct the test implementation independently of whichever provider ServiceLoader selects.

  - Retain Kafka’s compatibility fallback for applications using older framework versions, and document that decision.

  One qualification remains: deriving a POJO from a mutable cached map gives it that map’s values at conversion time. Later map mutations do not automatically update an already-cached POJO.
  The tests and documentation should state that boundary rather than promise continuous synchronization.

  The revised scope remains the cache fix, parsing safety, and regression coverage. No code changed.

stevehu:
Implemented and pushed to [PR #2780](https://github.com/networknt/light-4j/pull/2780) in 5f750d1a0.

  The repeated uncached-load workaround is replaced with stable representation caching, coordinated invalidation, and safe YAML parsing.

  Validation passed: cache regressions, clean config/handler/security/server suites, and all 65 Kafka tests. Both working trees are clean.

  Kafka validation against the published dependency remains pending framework publication.

stevehu:
Code review(high · 5 findings)
config/src/main/java/com/networknt/config/Config.java
  ● 686 [deadlock]        The new `synchronized (this)` guarding lazy ConfigLoader construction (lines 686 and 712) is reached from inside `cached()`'s `synchronized (entry.loadLock)` (line 176), so the order is loadLock -> this. The reverse order also exists: `ConfigLoaderConstructor.getInstance(...).getConfigLoader()` runs the user's ConfigLoader constructor while holding `this`, and any loader that reads configuration through the cached API (a config-server-backed loader reading its own settings is the obvious case) then acquires another entry's loadLock. Two threads loading two different config names deadlock permanently.
  ● 206 [error-handling]  The derived path calls `CentralizedManagement.convertMapToObj(map, type)` directly, bypassing `loadSpecificConfigFileAsObject`'s try/catch at line 500-503, which logs the exception and rethrows it wrapped as `RuntimeException("Unable to load <file> as object.", e)`. A Jackson binding failure now escapes as a bare `IllegalArgumentException` naming only the class, with no config file name and no log line.
  ● 200 [behavior-change] `if (type == Map.class) return cachedMap(...)` short-circuits before the configured-loader branch, so `getJsonObjectConfig(name, Map.class)` now resolves through `ConfigLoader.loadMapConfig` instead of `ConfigLoader.loadObjectConfig(name, Map.class)`. For any deployment with a `configLoaderClass` whose two methods return different content, the value silently changes.
  ● 165 [resource-leak]   `cacheEntry(name)` unconditionally installs a `CacheEntry` before the load is attempted, and `cached()` only stores a representation when the value is non-null (line 188). A config name that never resolves therefore leaves an empty CacheEntry in `configCache` forever, and it is only removed by `clear()`/`clearConfigCache`.
  ● 250 [behavior-change] `setClassLoader` gained a full `configCache.clear()`. That correctly drops cached `Class` references, but it also discards entries installed by `putInConfigCache`, which are not reloadable from the filesystem — they exist precisely because they came from somewhere else.

Findings

- config/.../Config.java:686 — [high] ABBA deadlock: the new synchronized (this) for lazy ConfigLoader init is nested inside cached()'s synchronized (entry.loadLock), while the loader's constructor can acquire another entry's loadLock. Two threads loading different config names hang permanently — ThreadMXBean.findDeadlockedThreads() reports 2 deadlocked threads on PR head, 0 on origin/master.
- config/.../Config.java:206 — [medium] The derived-POJO path calls convertMapToObj directly, bypassing loadSpecificConfigFileAsObject's logging and RuntimeException("Unable to load <file> as object.") wrapper. One unknown key now throws a bare IllegalArgumentException with no file name and no log line.
- config/.../Config.java:200 — [low] getJsonObjectConfig(name, Map.class) now routes to a custom ConfigLoader.loadMapConfig instead of loadObjectConfig; light-portal has six such call sites.
- config/.../Config.java:165 — [low] cacheEntry installs an entry before the load, so names that never resolve leave empty entries forever, and clearConfigCache now logs a cache eviction for configs that were never cached.
- config/.../Config.java:250 — [low] setClassLoader now clears the whole cache, silently dropping non-reloadable putInConfigCache injections (UrlConfigLoader:200); five existing tests call it on the process singleton.

Assessment

The design does what we discussed: identity is preserved per representation, invalidation is coordinated, getJsonObjectConfig(name, Map.class) shares the map entry, loader output is validated before insertion, and both YAML object-parsing branches are now locked. The derivation optimization is implemented with the right preconditions (configLoaderClass == null && !isExclusionConfigFile). Regression coverage is genuinely strong — 12 cache tests plus the two cross-module ones, including the "loading another name must not wait behind a blocked load" case.

Verified green: config 89 run / 5 skipped, server 16 / 2 skipped, status 21, info 2, handler-config 8. The security failures (72 run, 3 failures, 2 errors) reproduce identically on origin/master, so the PR's claim that they're pre-existing and out of scope is accurate — I confirmed that by reverting both changed main-source files and re-running.

Finding 1 is the only one I'd block on. It's a narrower scenario than it looks — it needs a configured configLoaderClass whose construction reads config through the cached API — but the failure mode is an unrecoverable startup hang, and it's a direct consequence of the lock structure this PR introduces. Worth noting master isn't clean here either: the same scenario StackOverflows there because the unsynchronized lazy init recurses. So the synchronized (this) is fixing something real; it just needs to live outside the entry lock. Constructing the loader eagerly in the AbstractConfigImpl constructor, or in a holder resolved before cached() is entered, removes the inversion without giving up the guard.

Environment restored: probes deleted, both working trees clean, patched config reinstalled to ~/.m2.

stevehu:
Addressed all five findings and pushed d144fa0e1 to [PR #2780](https://github.com/networknt/light-4j/pull/2780).

  Fixed loader-construction deadlock, binding-error context, custom Map dispatch, empty-entry cleanup, and injection preservation.

  All 19 cache regressions pass, along with clean security/server tests and all 65 Kafka tests. Both working trees are clean.

stevehu:
● Code review(high · 1 finding)
config/src/main/java/com/networknt/config/Config.java
  ● 221 [deadlock] The `configLoaderInitializingThread` bootstrap at line 180 covers only loader *construction*. Once the loader exists that field is null, so a `ConfigLoader.loadMapConfig`/`loadObjectConfig` implementation that reads a different config name runs on the normal path: it is invoked from `cached()` while holding `entry.loadLock` (line 221) and then acquires a second entry's loadLock through `withCacheEntry`. Two threads loading two names whose loads reference each other deadlock permanently.

Verification of d144fa0e1

Four of the five are fixed, verified by running against the code rather than reading it:

┌─────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│           Finding           │                                                                     Verified result                                                                      │
├─────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ #2 binding-error context    │ RuntimeException: Unable to load configuration 'errcfg' as object. — name present, cause chained, logger.error added                                     │
├─────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ #4 empty-entry accumulation │ 100 probes of missing config names and missing files leave configCache at size 1 (unchanged); one real load takes it to 2. Reader-count cleanup works,   │
│                             │ and clearConfigCache now only logs when representations exist                                                                                            │
├─────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ #5 injection preservation   │ Injection survives setClassLoader and is the same instance (sameInstance=true)                                                                           │
├─────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ #3 custom Map.class         │ Now gated on defaults || configLoaderClass == null, so a configured loader's loadObjectConfig is preserved when the object API is first reader           │
│ dispatch                    │                                                                                                                                                          │
├─────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ #1 loader-construction      │ Fixed — my original probe now completes on both threads, 0 deadlocked, vs. 2 deadlocked on 3868cc55d                                                     │
│ deadlock                    │                                                                                                                                                          │
└─────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

Suites: config 96 run / 0 failures / 5 skipped (19 in ConfigMixedRepresentationTest), plus status 21, info 2, handler-config 8, server 16 / 2 skipped — all green. Both working trees clean.

The configLoaderInitializingThread + detached-bootstrap-entry approach for #1 is a good solution: the initializing thread never touches a shared entry lock, so the cycle can't form, and bootstrap reads correctly avoid publishing default-loaded data as custom-loader data.

One residual finding

config/src/main/java/com/networknt/config/Config.java:221 — [high] The bootstrap guard covers loader construction only. A ConfigLoader.loadMapConfig/loadObjectConfig that reads a different config name runs under entry.loadLock and then takes a second entry's lock, so two threads loading mutually-referencing names deadlock. Verified with a depth-guarded, non-recursive probe: d144fa0e1 → both threads HUNG, 2 JVM-reported deadlocked threads; origin/master → both completed, 0 deadlocked.

I want to be clear that my first attempt at this probe was confounded — it recursed regardless of locking, which made master look equally broken. The isolated version above removes that: the loader is constructed up front, each thread cross-reads exactly once, and a barrier ensures both hold their own entry lock first. That version cleanly separates the two.

Two ways to close it, with a real trade-off:
- (a) Generalize the bootstrap trick from "this thread is constructing the loader" to "this thread is already inside a load," routing re-entrant reads to a detached CacheEntry. Preserves single-flight; keeps the existing tests passing.
- (b) Run the loader outside loadLock and publish via representations.putIfAbsent, returning the winner. Removes lock-held I/O entirely and still guarantees one identity per representation, but allows occasional duplicate loads — you'd need to relax concurrentLoadsConvergeOnOneIdentity's assertEquals(1, objectLoads.get()).

I'd take (a): it's a smaller change, it reuses a mechanism already in the codebase, and it keeps the single-flight guarantee the tests currently pin.

stevehu:
Resolved — and your 97 is right; my 101 was a counting artifact. The per-class JUnit 5 lines sum to 97 (96 at d144fa0e1, +1 for the new test), while surefire's summary line reads 101 because 4 tests run under the vintage provider print in a different format that my grep skipped. Same outcome either way: 0 failures, 0 errors, 5 skipped.

Verification of 3b4d7ef58

The approach-(a) fix is correct and I found nothing new.

The deadlock is closed for both callback shapes. With a depth-guarded, non-recursive cross-reader (loader constructed up front, barrier ensuring both threads hold their own entry lock before cross-reading):

CROSS_MAP    A=completed B=completed jvmDeadlocked=0
CROSS_OBJECT A=completed B=completed jvmDeadlocked=0

versus 2 JVM-reported deadlocked threads on d144fa0e1. I tested the object callback separately because loadObjectConfig reaches the entry lock by a different route than loadMapConfig; both are clean.

Top-level behavior is intact. getJsonObjectConfig returns a stable instance across calls, getJsonMapConfig likewise, and the shared entry still holds 2 representations — so derivation continues to publish the map to the real entry rather than to a detached one. That matters: cachedMap(entry, …) is called with the entry directly rather than through withCacheEntry, so the ThreadLocal correctly does not detach it. That's the subtle part of this fix and it's right.

Nested reads don't poison the identity mechanism. This was my one remaining concern — a config class whose load() first runs from inside a load would get a detached, uncached map, and the 50 classes using getMappedConfig() == mappedConfig would then see a mismatch. Verified it converges:

CONVERGE nestedWasDetached=true topLevelStable=true nestedValueCorrect=true

The nested read gets a correct value from a detached entry, and subsequent top-level reads are stable and identical. Worst case is one extra singleton rebuild, self-correcting — not a loop.

The ThreadLocal is handled correctly: saved and restored around loader.get(), remove()d when previously absent so it doesn't leak an entry, and restored on the exception path. Making it static is the right call — it also covers a callback that reads through getNoneDecryptedInstance(), which has its own separate cache.

The invariant now holds cleanly: a thread can hold at most one shared entry loadLock, since every shared acquisition goes through withCacheEntry and every nested one is detached. Combined with configLoaderLock being taken before any entry lock, and the configCache monitor never being held across one, there are no cycles left among these locks.

Suites: config 101 / 0 failures / 5 skipped (20 in ConfigMixedRepresentationTest), status 21, info 2, handler-config 8, server 16 / 2 skipped — all green. Both working trees clean, probes removed.

All six findings from this review are now addressed. The remaining pre-merge item is the one from the original review that isn't code: publish the patched light-4j config snapshot before qualifying light-kafka #258, since its regression test fails against the currently published 2.3.8-SNAPSHOT.

stevehu:
Thanks for verifying. All six code findings are resolved.

  The remaining sequence is:

  1. Merge light-4j PR #2780 and publish the patched 2.3.8-SNAPSHOT.
  2. Revalidate light-kafka PR #258 against that published snapshot, using a fresh dependency cache.
  3. Merge #258 once that validation passes.

  The local patched-artifact test does not replace the published-snapshot check.

PR #2781, opened 2026-09-10
[Bug Fix] add identity cached singleton load wherever missing in configs

stevehu:


stevehu:
Code review(high · 6 findings)
ingress-proxy/src/main/java/com/networknt/proxy/ExternalServiceConfig.java
  ● 231 [correctness] Changing the ModuleRegistry moduleClass from ExternalServiceHandler to ExternalServiceConfig orphans the only registration for the external-service config under the handler class name, breaking an existing test and targeted config reload.
ingress-proxy/src/main/java/com/networknt/proxy/mras/MrasConfig.java
  ● 154 [security]    MrasConfig.load() now registers mras.yml into ModuleRegistry with masks=null, publishing password fields through the server-info/module-registry endpoints; mras was previously not registered at all.
ingress-proxy/src/main/java/com/networknt/proxy/salesforce/SalesforceConfig.java
  ● 114 [security]    SalesforceConfig.load() registers salesforce.yml with masks=null, the same regression as MrasConfig: the config was previously unregistered and its intended mask list is in the commented-out SalesforceHandler.register().
router-config/src/main/java/com/networknt/router/middleware/TokenConfig.java
  ●  80 [concurrency] ModuleRegistry.registerModule mutates plain HashMap/ArrayList with no synchronization, and this PR makes it reachable from request-handling threads in several config classes that each hold a different class monitor.
router-config/src/main/java/com/networknt/router/RouterConfig.java
  ● 404 [correctness] RouterConfig.load(String) caches into the process-wide `instance` for any configName, unlike the nine other classes in this PR which only cache when configName equals CONFIG_NAME; switching the constructor to the cached map makes that pollution effective for the first time.
jaeger-tracing/pom.xml
  ●   8 [build]       The parent version bump to 2.3.8-SNAPSHOT is correct, but jaeger-tracing is absent from the root pom's <modules> list, so JaegerConfig's new singleton and the new ConfigReloadIdentityTest are never compiled or executed.

Reviewed PR #2781 (sync → master, 21 files). The core change is sound and fixes a real problem — several config classes had constructors calling getJsonMapConfigNoCache, which made the identity check instance.getMappedConfig() == cachedMap fail permanently, so RouterConfig.load() re-parsed YAML from disk on every request in RouterHandler.handleRequest. Six issues found; I built and ran the affected modules to verify.

Findings

- ingress-proxy/src/main/java/com/networknt/proxy/ExternalServiceConfig.java:231 — Registry key switched from ExternalServiceHandler to ExternalServiceConfig; ExternalServiceConfigTest:58 now fails (verified: expected: <true> but was: <false>), and since ExternalServiceHandler registers nowhere else, /config/reload targeting that handler class no longer finds external-service.
- ingress-proxy/src/main/java/com/networknt/proxy/mras/MrasConfig.java:154 — New registerModule(..., masks = null) publishes mras.yml (keyStorePass, keyPass, trustStorePass, basicAuth.password, accessToken.password) through /server/info; mras was previously unregistered and MrasHandler.register() (commented out) lists exactly those masks.
- ingress-proxy/src/main/java/com/networknt/proxy/salesforce/SalesforceConfig.java:114 — Same shape: certPassword now exposed unmasked. TableauConfig.java:85 is identical but tableau.yml has no secrets today.
- router-config/src/main/java/com/networknt/router/middleware/TokenConfig.java:80 — ModuleRegistry.registerModule mutates unsynchronized HashMap/ArrayList, now reachable from request threads: after a config reload, TokenConfig/ServiceDictConfig/PathPrefixServiceConfig each lock a different class monitor and register concurrently.
- router-config/src/main/java/com/networknt/router/RouterConfig.java:404 — The one class that didn't get the if (CONFIG_NAME.equals(configName)) guard its nine siblings got; a load("router-test") now permanently installs a non-default config as the shared instance, and alternating loads make RouterHandler rebuild its proxy per request.
- jaeger-tracing/pom.xml:8 — Parent version bump is correct, but jaeger-tracing isn't in the root <modules> (mvn -pl jaeger-tracing compile → "Could not find the selected project in the reactor"), so JaegerConfig and its new test are never compiled or run.

Everything else passed: service, restrans-config, router-config and ingress-proxy tests are green apart from the ExternalServiceConfigTest failure above (the security module failures are pre-existing and unrelated). No mutation of the now-shared cached maps in any of the changed setConfigData/setConfigList/setMap methods. The repo is back on its original detached HEAD at 9832db6b9.

stevehu:
Claude’s first three findings are confirmed; the remaining three do not establish newly introduced defects. Independent verification additionally found shared Salesforce token-state races,
  despite the five active new identity tests passing.

  Full review comments:

  - [P1] Mask MRAS credentials before registering the configuration — /home/steve/workspace/light-4j/ingress-proxy/src/main/java/com/networknt/proxy/mras/MrasConfig.java:154-154
    When MRAS credentials are configured, this new registration exposes them through server-info and module-registry responses even with maskConfigProperties=true. Passing null disables
    masking; using the non-decrypting loader does not protect plaintext credentials. Verified that trustStorePass remains readable. Supply masks covering keyStorePass, keyPass,
    trustStorePass, nested password fields, and microsoft.clientSecret.

  - [P1] Redact Salesforce secrets from the new registry entry — /home/steve/workspace/light-4j/ingress-proxy/src/main/java/com/networknt/proxy/salesforce/SalesforceConfig.java:114-114
    When Salesforce has a plaintext certificate password, this newly published configuration exposes certPassword through server-info and module-registry responses. Verified that the
    configured password remains readable with maskConfigProperties=true, because the mask argument is null. Redact certificate credentials and any password-grant secrets in pathPrefixAuths
    before publishing the configuration.

  - [P2] Preserve the external-service handler registration key — /home/steve/workspace/light-4j/ingress-proxy/src/main/java/com/networknt/proxy/ExternalServiceConfig.java:231-231
    Changing the registered class removes ExternalServiceHandler from the registry, so existing targeted reload requests naming that handler no longer clear the external-service cache. It
    also fails ExternalServiceConfigTest.testExternalServiceConfigRegisteredForServerInfo at line 58, reproduced against this PR. Preserve the existing registration identity or provide
    backward-compatible lookup for the old class name.

  - [P2] Synchronize token state before sharing Salesforce configurations — /home/steve/workspace/light-4j/ingress-proxy/src/main/java/com/networknt/proxy/salesforce/
    SalesforceConfig.java:105-106
    Concurrent requests for the same Salesforce prefix now share the same mutable PathPrefixAuth, whereas previously each load constructed fresh objects. SalesforceHandler.handleRequest()
    writes expiration before accessToken without synchronization. A second request between those writes sees a valid expiration, skips token acquisition, and sends Bearer null on initial
    acquisition or a stale token during refresh. Publish the token and expiration atomically and coordinate refreshes before enabling this shared-instance fast path.

stevehu:
The Preserve the external-service handler registration key is not necessary as this will only take effect for a new instance and there is no nagative impact at all. The only thing is
  changed is the /adm/server/info response and it is perfectly OK. So please remove this as an issue.

stevehu:
Dropped the external-service registration finding and updated its test expectation.

  Fixed the other three issues locally on sync:

  - MRAS and Salesforce credentials are masked, including JSON-string configurations.
  - Salesforce token refresh and publication are synchronized per authentication entry.

  Targeted Maven tests pass, including all nine new regression cases. Eight correctly fail against the original code. One existing test remains skipped.

  Changes are unstaged and uncommitted.

stevehu:
Both fixes verified. router.ConfigReloadIdentityTest is now 2/2 (my new test passes), all com.networknt.proxy tests green, BUILD SUCCESS. The only failures in the reactor are the pre-existing security module JwtVerifier* ones, which were already failing on the untouched PR branch in round 1.

---

PR #2781 review — round 2 handoff for Codex

Context

networknt/light-4j#2781, branch sync → master: "[Bug Fix] add identity cached singleton load wherever
missing in configs".

The PR gives ten config classes an identity-cached singleton load() and switches their constructors from
getJsonMapConfigNoCache to getJsonMapConfig. This fixes a real and serious problem: with a NoCache
constructor the identity check instance.getMappedConfig() == Config.getInstance().getJsonMapConfig(name)
could never succeed, so load() re-parsed YAML from disk on every call — and RouterHandler,
TokenHandler, SalesforceHandler, PathPrefixServiceHandler, ServiceDictHandler all call load() per
request.

Verification

mvn -o -pl router-config,ingress-proxy -am test → BUILD SUCCESS.

router.RouterConfigTest                     9/9
router.ConfigReloadIdentityTest             2/2   (was 1, +onlyDefaultConfigNameIsCached)
router.middleware.*ConfigTest               5/5
proxy.CredentialRegistryMaskingTest         4/4   (new)
proxy.salesforce.SalesforceTokenConcurrencyTest  5/5   (new)
proxy.ExternalServiceConfigTest             5/5   (was 4/5)
proxy.ConfigReloadIdentityTest              1/1

Pre-existing and unrelated: security module JwtVerifierTest / JwtVerifierMultipleJwkTest (5 failures,
expired fixtures). Present before any change in this PR.

Round 1 outcomes

┌───────────────────────────────┬───────────────────────────────────────────────────────────────────┐
│            Finding            │                              Outcome                              │
├───────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│ MrasConfig:154 secrets        │ Fixed by Codex. I re-checked mras.yml field by field — coverage   │
│ unmasked                      │ is complete (anonymous holds only serviceHost).                   │
├───────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│ SalesforceConfig:114          │ Fixed by Codex. Complete against salesforce.yml.                  │
│ certPassword unmasked         │                                                                   │
├───────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│ SalesforceHandler shared      │ Fixed by Codex. Codex's find, I missed it. Genuinely new: pre-PR  │
│ PathPrefixAuth race           │ each request got fresh PathPrefixAuth objects (so the token cache │
│                               │  never worked at all); post-PR they're shared.                    │
├───────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│ ExternalServiceConfig:231     │ Dropped per maintainer. Only affects /adm/server/info;            │
│ registry key                  │ ExternalServiceHandler never registered itself. Test expectation  │
│                               │ updated.                                                          │
├───────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│                               │ Withdrawn — my error. Pre-PR RouterHandler drove registerModule   │
│ ModuleRegistry HashMap race   │ on every request from every worker thread; this PR removes that,  │
│                               │ so it's a large net improvement. ModuleRegistry's collections     │
│                               │ remain unsafe — separate issue.                                   │
├───────────────────────────────┼───────────────────────────────────────────────────────────────────┤
│                               │ Accepted per maintainer. Module is deliberately out of the        │
│ jaeger-tracing/pom.xml        │ reactor pending an OpenTelemetry rewrite; the JaegerConfig change │
│                               │  stays so the old pattern isn't reintroduced.                     │
└───────────────────────────────┴───────────────────────────────────────────────────────────────────┘

Changes made this round

1. router-config/src/main/java/com/networknt/router/RouterConfig.java:405 — added the missing
CONFIG_NAME guard

RouterConfig was the only one of the ten without it. Pre-PR the omission was masked because the NoCache
constructor made every load() rebuild and overwrite instance immediately. Post-PR the pollution is
durable: one RouterConfig.load("anything-else") pins the process-wide instance to a non-default config,
and registers a second moduleClasses entry so
ConfigReloadHandler.findConfigName("com.networknt.router.RouterConfig") can return the wrong config name
and clear the wrong cache. Uses the same early-return shape as ExternalServiceConfig.load. No in-repo
caller passes a non-default name — hardening, not a live bug fix.

2. ingress-proxy/src/main/java/com/networknt/proxy/mras/MrasHandler.java — the same token race Codex
fixed in Salesforce

MrasHandler is a chain singleton, so accessToken/accessTokenExpiration and microsoft/microsoftExpiration
are shared across worker threads, non-volatile and unsynchronized, and both wrote the expiration before
the token. A thread arriving between those two writes sees a valid expiration, skips acquisition, and
sends Bearer null on first acquisition or a stale token on refresh. Fixed with the same shape as
SalesforceHandler: per-cache lock (accessTokenLock, microsoftLock), token published before expiration,
token read back inside the lock into a local, null-token guard, 1000L.

This one is pre-existing, not introduced by the PR — included at the maintainer's request because it's
the identical defect one file away.

3. router-config/src/test/java/com/networknt/router/ConfigReloadIdentityTest.java —
onlyDefaultConfigNameIsCached, mirroring ExternalServiceConfigTest.testOnlyDefaultConfigNameIsCached.
Order-independent w.r.t. the existing test in the same class.

Open items for Codex

- MrasHandler has no coverage for the fix. Both MrasHandlerTest methods are @Disabled (they need real MRAS credentials and two-way TLS). Unlike SalesforceHandler — whose client field is injectable, which is what made SalesforceTokenConcurrencyTest possible — MrasHandler.getAccessToken builds its HttpClient inline via createSSLContext(config). The fix is verified by inspection and compilation only. Making it testable means extracting the client the way SalesforceHandler has it; judgement call whether that belongs in this PR.
- oauthServer.yml client_credentials is published unmasked (a list of client_id:client_secret). Not a regression — OAuthServerHandler:47 already registered it with null masks pre-PR, and OAuthServerConfig.load() now adds a duplicate entry. Note ModuleRegistry can't fully fix it: maskList compares the element value to the mask string, so only the JSON-string form is maskable by key name; the YAML-list form can't be masked at all.
- Checked, clean, no change made: TokenHandler's JWT cache builds a fresh Jwt locally and only publishes it into the ConcurrentHashMap after success, so there is no half-published state (it permits duplicate concurrent fetches but never serves a bad token). TableauConfig and ExternalServiceConfig register with null masks but their YAML carries no secrets.

Working tree (unstaged, uncommitted on sync)

M ingress-proxy/.../proxy/mras/MrasConfig.java              (Codex: masks)
M ingress-proxy/.../proxy/mras/MrasHandler.java             (me: token race)
M ingress-proxy/.../proxy/salesforce/SalesforceConfig.java  (Codex: masks)
M ingress-proxy/.../proxy/salesforce/SalesforceHandler.java (Codex: token race)
M ingress-proxy/.../test/proxy/ExternalServiceConfigTest.java (Codex: key expectation)
M router-config/.../router/RouterConfig.java                (me: CONFIG_NAME guard)
M router-config/.../test/router/ConfigReloadIdentityTest.java (me: +1 test)
? ingress-proxy/.../test/proxy/CredentialRegistryMaskingTest.java        (Codex, new)
? ingress-proxy/.../test/proxy/salesforce/SalesforceTokenConcurrencyTest.java (Codex, new)

stevehu:
One remaining PR regression is confirmed:

  - [P2] Initialize the Salesforce HTTP client independently of token acquisition — light-4j/ingress-proxy/src/main/java/com/networknt/proxy/salesforce/SalesforceHandler.java:204. Two handler
    instances sharing the default configuration also share its token, but each owns its HTTP client. After the first acquires a token, the second skips acquisition—and client initialization—
    then fails in invokeApi() with this.client == null. Reproduced HTTP 500 on current code versus HTTP 200 with the pre-PR SalesforceConfig, without client injection. Add a two-instance
    regression test alongside the fix.

  The RouterConfig guard and MRAS locking changes check out. MRAS tests need no production refactor: a scratch probe using local HTTP endpoints and a test keystore passed initial acquisition
  and refresh for both token types, each with eight concurrent requests and one acquisition.

  The focused offline Maven run passed: 46 passed, eight existing skips. OAuth credential masking remains a separate follow-up. No repository files changed during this review.

stevehu:
Fixed locally on sync. Each Salesforce handler now safely initializes its HTTP client before checking the shared token.

  Both new two-handler regression cases failed before the fix and pass now. The focused router/proxy build passed: 48 tests passed, eight existing skips.

  Changes remain unstaged and uncommitted.

stevehu:
Fix reviewed and approved on inspection. Final test run is in flight — I'll report the numbers as soon as the monitor fires.

Summary of where things stand:

Codex's SalesforceHandler fix is correct. It addressed all three concerns I raised: volatile field plus a synchronized double-checked initializeClient, init placed inside the matched-prefix branch (so no SSL context work for non-Salesforce traffic), and the two duplicated blocks collapsed into one helper with the error path converted to setExchangeStatus. I traced every client.send site and confirmed all three are reachable only downstream of the new init, and checked the two locks can't nest.

Final state of the branch — 7 modified, 2 new test files:

┌───────────────────────────────────────────────────────────┬────────┬─────────────────────────────────────────────┐
│                           File                            │ Author │                   Change                    │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ proxy/mras/MrasConfig.java                                │ Codex  │ mask credentials on registration            │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ proxy/mras/MrasHandler.java                               │ me     │ token race (pre-existing, fixed on request) │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ proxy/salesforce/SalesforceConfig.java                    │ Codex  │ mask credentials on registration            │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ proxy/salesforce/SalesforceHandler.java                   │ Codex  │ token race + client init regression         │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ test/proxy/ExternalServiceConfigTest.java                 │ Codex  │ registry key expectation                    │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ router/RouterConfig.java                                  │ me     │ CONFIG_NAME guard                           │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ test/router/ConfigReloadIdentityTest.java                 │ me     │ onlyDefaultConfigNameIsCached               │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ test/proxy/CredentialRegistryMaskingTest.java             │ Codex  │ new                                         │
├───────────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────┤
│ test/proxy/salesforce/SalesforceTokenConcurrencyTest.java │ Codex  │ new                                         │
└───────────────────────────────────────────────────────────┴────────┴─────────────────────────────────────────────┘

Two items deliberately left open, both documented and neither a regression from this PR: oauthServer.yml's client_credentials published unmasked (pre-existing via OAuthServerHandler:47, and only partly fixable given how maskList matches), and ModuleRegistry's unsynchronized collections (which this PR substantially improves rather than worsens).

Everything remains unstaged and uncommitted on sync.

PR #2786, opened 2026-09-10
Restore deprecated Config Server lookup parameters for backward compatibility

### LINKED ISSUE #2785: Restore deprecated Config Server lookup parameters for client backward compatibility
## Problem

The client changes in #2777 broke compatibility with a customer-provided Config Server that depends on the legacy configuration lookup parameters.

- Commit `94e87fb3f` removed forwarding of `productId`, `productVersion`, `apiId`, and `apiVersion` from `DefaultConfigLoader`.
- Commit `304694612` made `serviceId` mandatory. Restoring query parameters alone would not restore compatibility for legacy configurations that omit `serviceId`.

Retain the existing client contract for custom Config Servers while deprecating the legacy parameters. New deployments should use published configuration snapshots instead of live-database configuration selection.

## Scope

Changes belong in light-4j's client bootstrap code, `server/src/main/java/com/networknt/server/DefaultConfigLoader.java`, its tests, and `server/src/main/resources/config/startup.yml`.

**No changes are needed in our Config Server.** Its lookup identity remains `host`, `serviceId`, and `envTag`; its snapshot-based behavior must remain unchanged. Forwarding legacy parameters to a customer's server does not dictate how that server obtains configuration.

## Proposed behavior

- Restore forwarding of `productId`, `productVersion`, `apiId`, and `apiVersion` only when explicitly configured, preserving the historical parameter names and values.
- Existing legacy startup configuration should work without requiring a new compatibility flag.
- Keep snapshot identity validation for normal configurations. When legacy parameters are configured, support the historical request shape, including an omitted `serviceId`.
- Preserve the current host/environment defaults and unrelated bootstrap behavior; implement a targeted compatibility restoration rather than reverting the entire change.
- Mark legacy lookup parameters as deprecated in code comments and commented `startup.yml` examples. Explain that they are retained for custom Config Server compatibility and that new deployments should use `host`, `serviceId`, and `envTag` with published snapshots.
- Emit one deprecation warning during bootstrap when legacy parameters are used, rather than repeating it for each configuration, certificate, or file request.

## Acceptance criteria

- [ ] Normal snapshot requests continue to use `host`, `serviceId`, and `envTag`, without adding unconfigured legacy parameters.
- [ ] Normal configurations still reject a missing or blank `serviceId`.
- [ ] Explicitly configured legacy parameters are forwarded with their original names and values.
- [ ] Legacy configuration without `serviceId` can reach a custom Config Server.
- [ ] Compatibility applies consistently to configuration, certificate, and file downloads that use the shared query builder.
- [ ] Bootstrap emits one deprecation warning for legacy usage and none for normal snapshot usage.
- [ ] Code and `startup.yml` document deprecation and the recommended snapshot contract.
- [ ] Targeted regression tests cover both request formats and validation behavior.
- [ ] Our Config Server implementation and snapshot lookup contract remain unchanged.

stevehu:
## Summary

Restore compatibility with customer-provided Config Servers after #2777: `DefaultConfigLoader` forwards nonblank configured `productId`, `productVersion`, `apiId`, and `apiVersion`, and allows legacy requests to omit `serviceId`. No compatibility flag is required. This automatic detection is intentional for existing customer deployments. Normal snapshot requests still require `serviceId`; host/environment validation remains in place.

Null, empty, and whitespace-only legacy values are ignored and do not enable compatibility mode or trigger a warning. Every query value is URL-encoded; nonblank legacy values are trimmed before encoding, preserving internal spaces and delimiters after decoding. Blank serviceId values are omitted from legacy requests. Startup builds the query once and reuses it for configurations, certificates, and files.

Code and `startup.yml` document deprecation and one warning per bootstrap/reload. The portal-service Config Server uses only `host`, `serviceId`, and `envTag` for snapshot lookup. The deprecated Java light-config-server retains its legacy behavior. Neither server implementation is changed.

## Validation

- `mvn -pl server -am -Dtest=DefaultConfigLoaderTest,LegacyConfigServerQueryTest -Dsurefire.failIfNoSpecifiedTests=false test` — passed, 16 tests, no failures or skips.
- Coverage includes all four legacy parameters with/without `serviceId`, individual fields, null/blank omission, snapshot identity validation, and warning behavior.
- Query round-trip coverage includes spaces, Unicode, and URI delimiters across configurations, certificates, and files URL paths.
- `git diff --cached --check` passed; staged GitNexus analysis reports low risk. Actual diff is limited to client query construction, documentation, and tests.

Fixes #2785


stevehu:
Code review(high · 4 findings)
light-4j/server/src/main/java/com/networknt/server/DefaultConfigLoader.java
  ● 548 [correctness] Forwarding productId and productVersion is not inert against networknt's own Config Server — it switches /configs, /certs and /files from published-snapshot lookup to live-instance (database) lookup, contradicting the premise of the PR and of the new startup.yml comment.
  ● 536 [correctness] Legacy mode is inferred from the mere presence of any of the four keys, so a stale productId/apiId left in an existing startup.yml silently disables the serviceId requirement added in 304694612 for deployments that are otherwise normal snapshot clients.
  ● 537 [correctness] The legacy detection and the emission loop test `!= null` while host/serviceId/envTag in the same method are trimmed and blank-checked, so an empty-string legacy value is treated as an explicit legacy configuration.
  ● 550 [correctness] Legacy values are appended raw while host/serviceId/envTag are trimmed, so a value containing whitespace or a URI delimiter produces an illegal query string that aborts bootstrap before the local-cache fallback can run.

I reviewed PR #2786 against issues #2785 and #2777, and cross-checked the actual light-config-server implementation in your workspace. The build is clean: mvn -pl server -am -Dtest=DefaultConfigLoaderTest,LegacyConfigServerQueryTest test passes, 13 tests, 0 failures — matching the PR's claim.

Findings (most severe first):

- light-4j/server/src/main/java/com/networknt/server/DefaultConfigLoader.java:548 — Forwarding productId+productVersion is not inert against your own Config Server: ConfigsGetHandler/CertsGetHandler/FilesGetHandler branch on hasText(productId) && hasText(productVersion) → live-instance lookup, else snapshot lookup. Any existing deployment that still carries those keys (they were required before #2777) silently reverts from published snapshots to live-database config on upgrade.
- DefaultConfigLoader.java:536 — Legacy mode is detected from the presence of any of the four keys, so a stale productId/apiId disables the serviceId guard from 304694612. A missing or mistyped serviceId no longer fails fast; the request goes to the server's else branch where productId is treated as serviceId and the token sid authorization check is skipped entirely.
- DefaultConfigLoader.java:537 — Blank legacy values count as "configured" (!= null vs. the trimmed blank checks used for host/serviceId/envTag), so productId: "" from a values.yml/env injection drops the serviceId guard, appends &productId=, and logs a spurious deprecation warning.
- DefaultConfigLoader.java:550 — Legacy values are appended untrimmed and unencoded; a value with whitespace makes URI.create throw outside getServiceConfigs's try block, aborting bootstrap and bypassing the local values.yml fallback.

The first two undercut the PR's central premise ("Our Config Server ... continues to use only host, serviceId, and envTag"), and the same wrong claim is now baked into startup.yml. If the intent really is that legacy params only reach customer servers, the loader needs to gate them on a custom-server signal (or a flag), not just on their presence.

I also noted, outside the diff's scope: light-config-server's ServiceIdAuthorizationUtil.authorizeRequestContext skips the sid claim check whenever serviceId is absent — this PR makes that request shape reachable again from light-4j clients. Worth a separate issue on the server side.

stevehu:
Updated [PR #2786](https://github.com/networknt/light-4j/pull/2786):

  - Blank legacy values are ignored.
  - Query values are URL-encoded.
  - Documentation distinguishes the server implementations.

  The compatibility tradeoff is preserved; neither server was changed. All 15 targeted tests passed.

stevehu:
Re-reviewed commit 379965850. Findings 3 and 4 are properly fixed, and I verified the surrounding claims:

Verified
- Blank handling: isBlank() on the raw value now drives both legacy detection and emission, so an empty-string legacy value no longer bypasses the serviceId guard, no longer appends an empty parameter, and no longer logs a spurious deprecation warning. The new test covers "", " ", "\t\n" and U+2003 across all four keys.
- Encoding: URLEncoder.encode(..., UTF_8) on every value closes the illegal-URI crash and the &/= injection path; the round-trip test asserts no duplicate keys and no fragment across all three endpoints. +-for-space is safe here — Undertow parses query strings with formEncoding=true (URLUtils.java:300), so light-4j-based servers decode it correctly. Normal snapshot values (dev.lightapi.net, com.networknt.example-1.0.0, dev) pass through the encoder unchanged, so there's no wire-format change for the standard path.
- Objects.toString on legacy values also fixes non-String YAML values (e.g. an unquoted version parsed as a number) that previously went through append(Object).
- Doc correction is accurate: portal-service/apps/config-server genuinely ignores the legacy fields (its own test config_query_ignores_removed_product_identity_fields asserts it), and it can resolve serviceId from claims when omitted.
- Tests: full server module passes (LegacyConfigServerQueryTest now 7, DefaultConfigLoaderTest 8). The reactor's 3 failures / 2 errors are all in the security module (JwtVerifierTest, JwtVerifierMultipleJwkTest) and reproduce identically on master — pre-existing, unrelated.

Two leftovers, both in the same blank/trim family:

- light-4j/server/src/main/java/com/networknt/server/DefaultConfigLoader.java:552 — the serviceId gate is still startupConfig.get(SERVICE_ID) != null while legacy params moved to isBlank(); a legacy config with a blank serviceId emits &serviceId=, so a custom server that branches on parameter presence does an empty-string lookup instead of the productId-only legacy path. if (!serviceId.isEmpty()) makes it consistent and doesn't change the normal path.
- DefaultConfigLoader.java:539 — legacy values are blank-checked but not trimmed, unlike host/serviceId/envTag; productId: "light-4j " now goes out as productId=light-4j+ and silently mismatches on the server where it previously failed loudly. Storing value.trim() finishes the fix; the round-trip test currently asserts padding is preserved, so it would need updating.

Both are minor and optional — nothing else in the diff looks wrong to me.

stevehu:
Updated [PR #2786](https://github.com/networknt/light-4j/pull/2786) with commit bf03bf1d7:

  - Blank serviceId is omitted from legacy requests.
  - Nonblank legacy values are trimmed before encoding.
  - Tests and documentation updated.

  All 16 targeted tests passed.

PR #2787, opened 2026-09-13
fixes #2235 match handler.yml paths with or without a trailing slash

### LINKED ISSUE #2235: evaluate if we can support the path matching in handler.yml with or without trailing slash
```
ie:

 Handler.yml:

- path /foo/v1
  - com.networknt.FooHandler

API consumer requests /foo/v1/

As it is, this fails to match the path, and so we have had to populate our handler.yml with dual path specs for every API to allow consumers to call our APIs with or without trailing slashes.  This is a nuisance and it would be great if we could auto-match these.

```

gohri81-tech:
## Summary

`Handler.start()` matched the request path against the configured `handler.yml` paths with `PathTemplateMatcher.match()` only. The matcher normalizes an empty path and a missing leading slash, but not a trailing one, so a request to `/foo/v1/` does not match a path configured as `/foo/v1` and every path has to be declared twice.

`Handler.matchPath()` now retries the match with the trailing slashes removed when the exact match fails. An exact match always wins, and the fallback is confined to the `handler.yml` chain lookup instead of the shared `PathTemplateMatcher`.

## Opt in

The fallback is off by default and is turned on with the new `trailingSlashFallback` flag in `handler.yml`, so the chain that an existing deployment resolves for a request does not change on upgrade and every deployment decides for itself.

When the fallback matches, `normalizeRequestPath()` rewrites the request path, the relative path and the request URI of the exchange to the matched path, so that the chain that runs and the path forwarded downstream cannot disagree. The URI is trimmed on its own terms rather than by cutting the decoded request path off its end: `trimTrailingSlashesFromURI()` works on the path part of the raw URI, which ends at the first semicolon, so a URI carrying path parameters like `/foo/v1/;x=1` keeps them. When the path part does not end with the slashes that were trimmed from the request path, which is the case for a percent encoded path such as `/caf%C3%A9/`, the URI cannot be made to agree with the chain, so the fallback match is dropped and the request falls through to `defaultHandlers` as it does today. Every trailing slash is trimmed, so `/foo/v1//` resolves to `/foo/v1` as well, while the root path is left alone. A fallback match is logged at info, a skipped one at warn.

## Notes

The opposite direction (a request without a trailing slash matching a path configured as `/foo/v1/`) is deliberately left out, because it would let `/foo` match `/foo/{id}` with an empty path parameter. Happy to add it if you would rather have the symmetric behaviour.

I am happy to follow up with a light-doc change describing the flag if this direction looks right.

## Tests

`HandlerTest` covers the disabled default, the enabled fallback with multiple trailing slashes including the normalized exchange, the encoded URI that has to be skipped, the URI with path parameters, and the trimming edge cases. I was not able to run the Maven build in my environment, so please let CI be the judge on that.

Fixes #2235


stevehu:
Code review(low · 1 finding)
handler/src/main/java/com/networknt/handler/Handler.java
  ● 413 [security] The fallback sends `/x/` requests to the `/x` chain (or its endpoint security, rate-limit, or whitelist config) instead of `defaultHandlers`, but the exchange's request path still ends in `/`, so downstream proxy or router handlers forward `/x/` unchanged.

I found one possible problem in PR #2787, marked plausible rather than confirmed. Otherwise the change looks correct.

- handler/src/main/java/com/networknt/handler/Handler.java:413 — the chain changes but the forwarded path doesn't.
  - Before: a request like /public/ found no match and went to defaultHandlers.
  - After: it runs the /public chain, with that path's own security settings. The request path still ends in /, so the proxy sends /public/ to the backend unchanged.
  - Why it matters: a gateway may have an unauthenticated /public chain and an authenticated default chain. Then /public/ goes from requiring auth to not requiring it. A backend that treats /public/ as a different resource from /public would then serve that resource without auth.
  - Possible fix: since the PR adds no opt-in switch, existing deployments get this automatically. Either make the fallback opt-in from handler.yml, or also strip the trailing slash from the request path when the fallback matches.

What I checked in the matcher and path template code:
- Parameterized paths: templates like /foo/{id} already match /foo/bar/ on the first try. The retry only affects plain paths, which need an exact length match, so it doesn't change which chain existing requests get.
- Edge cases: the root path, empty or null paths, and wildcard paths are all handled.
- Tests: they use the right setup. handler.yml has GET /test, and the matcher map is visible to the test. I didn't run them.

stevehu:
Following up on the trailing-slash fallback finding — thinking about it more, I'd push this from "plausible" toward a real blocker, for a reason beyond the single /public vs /public/ example: the fallback is unconditional. There's no handler.yml switch to opt out, so every existing deployment inherits this new matching behavior on upgrade with no way to disable it. Any config where a trailing-slash request currently falls through to defaultHandlers (or a different chain) will start resolving to a different chain's security/rate-limit/whitelist rules the moment someone upgrades — silently, since the only logging is at trace.

Two more things worth resolving before merge:

1. Path/chain mismatch: matchPath() picks the chain based on the trimmed path, but ex.getRequestPath() itself is never updated. So a request to /foo/v1/ runs the /foo/v1 chain's security config, but if that chain proxies downstream, the backend still receives /foo/v1/ unchanged. If a backend treats those as distinct resources, the chain's auth decision and the actual forwarded resource disagree.
2. Only one trailing slash is trimmed: /foo/v1// won't match /foo/v1, so the fix doesn't fully generalize to "ignore trailing slashes" — worth knowing so it isn't assumed to be a complete fix for #2235.

Given the two directions you offered in the PR description, I'd lean toward making this opt-in from handler.yml — that avoids the upgrade blast-radius issue entirely and lets each deployment decide. If you'd rather keep it default-on, then normalizing ex.getRequestPath() when the fallback matches (so chain and forwarded path stay consistent) plus bumping the log to info/warn would address the main risk.

gohri81-tech:
Thanks for the detailed follow-up — I've taken the opt-in direction you leaned towards, and also fixed the other two points so none of them are left open.

handler.yml gets a new trailingSlashFallback flag, default false. matchPath() only retries the match when it is enabled, so upgrading changes nothing for an existing deployment and each one can decide for itself. When the fallback does match, normalizeRequestPath() now rewrites the request path, the relative path and the request URI of the exchange to the matched path, so the chain that runs and the path forwarded downstream can no longer disagree. The request URI is only rewritten when it ends with the request path, which leaves an encoded URI alone. trimTrailingSlash is now trimTrailingSlashes and trims all of them, so /foo/v1// resolves to /foo/v1 as well, with the root path still untouched. The match is logged at info rather than trace.

Tests cover the disabled default, the enabled path with multiple slashes including the normalised exchange, and the trimming edge cases. I still can't run the Maven build locally, so please let CI be the judge.

stevehu:
Code review(low · 1 finding)
handler/src/main/java/com/networknt/handler/Handler.java
  ● 464 [security] normalizeRequestPath only rewrites the request URI when it literally ends with the decoded request path, but ProxyHandler.createProxyRequestTargetURI builds the backend path from getRequestURI(), so the chain/forwarded-path mismatch from the earlier finding still happens for percent-encoded (or ;path-param) URIs.

The latest changes fix all three points from last time. One edge case is still open.

What's fixed:
- Upgrade risk: the new trailingSlashFallback flag is off by default. It's defined in handler.yaml, the schema and HandlerConfig, and matchPath() only retries when the flag is on.
- Matched chain vs. forwarded path: when the fallback matches, the request path, the relative path and the request URI are rewritten to the matched path. A match is now logged at info.
- Multiple slashes: trimTrailingSlashes removes all of them, so /foo/v1// works, and it leaves the root path / alone.
- Tests: they cover the default being off, the fallback with the rewritten request, and the trimming edge cases. I didn't run them, and CI results weren't part of this check.

Still open:
- handler/src/main/java/com/networknt/handler/Handler.java:464 (plausible, not confirmed): the request URI is only rewritten when it ends with the request path. Undertow decodes the path but not the URI, so a percent-encoded request like /caf%C3%A9/ skips the URI rewrite, and so does a URI with a ;param part like /public/;x=1. ProxyHandler.createProxyRequestTargetURI() builds the backend path from getRequestURI(). The /café chain's security therefore runs, but the backend still gets /caf%C3%A9/. This is the same mismatch as before, now limited to encoded or ;param URIs.
  - Possible fix: trim the trailing slashes from the raw URI itself (before any ;), or skip the fallback when the URI can't be rewritten to match.

It's a narrow case, since it only affects plain configured paths that contain encoded characters or requests that carry ; parameters. Given the auth point last time, it's worth closing or at least documenting before merge.

gohri81-tech:
Thanks - that last one is closed too now, in 9e3574f with tests in 798f051.

You were right that the check was the wrong shape. Cutting the decoded request path off the end of the URI can only work when the two agree character for character, which is exactly what an encoded path or a `;` parameter breaks. The URI is now trimmed on its own terms instead: `trimTrailingSlashesFromURI` takes the path part of the raw URI, which ends at the first semicolon, and removes as many trailing slashes from it as were removed from the request path, then puts the parameters back. So `/public/;x=1` becomes `/public;x=1` and `ProxyHandler.createProxyRequestTargetURI` builds the backend path from a URI that agrees with the chain.

For the case where that is not possible, `/caf%C3%A9/` being the obvious one, I took your second option rather than trying to be clever with the encoding. `normalizeRequestPath` now returns false when the path part does not end with the slashes that were trimmed, and `matchPath` throws the fallback match away when that happens, logging at warn. The request then falls through to `defaultHandlers` exactly as it does today, so there is no path left where the chain that runs and the resource that is forwarded can disagree. Erring towards no match keeps the security decision on the safe side, and the whole feature stays behind `trailingSlashFallback` anyway.

Tests cover both directions through `matchPath` (encoded URI skipped and the exchange left untouched, `;x=1` matched and normalized) plus `trimTrailingSlashesFromURI` on its own, including the absolute `http://localhost/test/` form. As before I can't run the Maven build here, so please treat CI and the Sonar gate as the source of truth.

stevehu:
The two new commits (9e3574f and 798f051) have no correctness findings, and they resolve the finding from the last review.

What the new commits fix
- Encoded URIs are handled safely. trimTrailingSlashesFromURI now trims the raw URI itself instead of cutting the decoded path off its end. If the raw path doesn't end in the slashes that were trimmed (for example /caf%C3%A9/ or /test%2F), matchPath drops the fallback match. The request then goes to defaultHandlers, so the chain that runs and the forwarded path can't disagree.
- The exchange is only changed on success. The normalized URI is computed before any setter runs, so when the fallback is rejected the exchange is left exactly as it was. The test at HandlerTest.java:298-299 checks this.
- ; path parameters are kept. I checked this against Undertow 2.4.1 Connectors.setExchangeRequestPath: it strips ;params from requestPath but leaves them in requestURI. That is the case the new code handles, so /test/;x=1 becomes /test;x=1.

One small point, not a finding. trimTrailingSlashesFromURI splits the URI at the first ;, but the slashes being trimmed come before the last segment's parameters. Take /a/;x=1/b/: the request path is /a//b/, and the code would produce /a;x=1/b/. The code only reaches this if a configured template matches /a//b but not /a//b/. com.networknt.utility.PathTemplateMatcher doesn't normalize slashes, so I don't think that can happen in practice. If you want to harden it anyway, look for the ; only after the last /: requestURI.indexOf(';', requestURI.lastIndexOf('/')).

Sonar Quality Gate issue
The new issue is java:S3776 at Handler.java:408: "Refactor this method to reduce its Cognitive Complexity from 22 to the 15 allowed." The new if/else in matchPath added two more nested log guards. Two changes fix it:
1. Move the fallback into its own method and use early returns instead of nesting.
2. Drop the isInfoEnabled/isWarnEnabled guards. SLF4J already skips building the message when the level is off, and the arguments are plain strings or a cheap getter.

static PathTemplateMatcher.PathMatchResult<String> matchPath(PathTemplateMatcher<String> pathTemplateMatcher, HttpServerExchange ex) {
    final var requestPath = ex.getRequestPath();
    final var result = pathTemplateMatcher.match(requestPath);

    if (result != null || config == null || !config.isTrailingSlashFallback())
        return result;

    return matchWithoutTrailingSlashes(pathTemplateMatcher, ex, requestPath);
}

private static PathTemplateMatcher.PathMatchResult<String> matchWithoutTrailingSlashes(PathTemplateMatcher<String> pathTemplateMatcher, HttpServerExchange ex, String requestPath) {
    final var trimmedPath = trimTrailingSlashes(requestPath);

    if (trimmedPath == null)
        return null;

    final var result = pathTemplateMatcher.match(trimmedPath);

    if (result == null)
        return null;

    // keep the chain and the path that is forwarded downstream consistent with each other.
    if (!normalizeRequestPath(ex, requestPath, trimmedPath)) {
        LOG.warn("Request path {} is matched to the configured path {} after the trailing slashes are removed, but the request URI {} cannot be normalized to the matched path, so the fallback is skipped.", requestPath, trimmedPath, ex.getRequestURI());
        return null;
    }

    LOG.info("Request path {} is matched to the configured path {} after the trailing slashes are removed. The request path of the exchange is normalized to the matched path.", requestPath, trimmedPath);
    return result;
}

By Sonar's scoring, this brings matchPath down to about 2 and the helper to about 3. Behavior and the existing tests stay the same.

Separately, Sonar shows 0.0% coverage on new code even though the PR adds tests. Coverage from the handler module probably isn't being reported to Sonar. The gate still passed, but that's worth checking.

gohri81-tech:
Thanks - both of the remaining points are in 3c1debc.

**Sonar S3776 at `Handler.java:408`.** Taken as you wrote it. `matchPath` is now the exact match, the two guards and a delegation, and the fallback lives in `matchWithoutTrailingSlashes` with early returns instead of four levels of nesting. The `isInfoEnabled` and `isWarnEnabled` guards are gone, since SLF4J skips building the message when the level is off and the arguments are plain strings and a cheap getter. Behaviour is unchanged, so `HandlerTest` needed no edit.

**The `;` before the last `/`.** Hardened as well, even though I agree it is not reachable today: `trimTrailingSlashesFromURI` now uses `requestURI.indexOf(';', requestURI.lastIndexOf('/'))`, so a parameter on an earlier segment, as in `/a/;x=1/b/`, stays where it is rather than being treated as the end of the path part. `lastIndexOf` returning `-1` behaves like a search from `0`, so a URI with no slash at all is unaffected. The javadoc says "the first semicolon that follows the last slash" now, and `/test/;x=1` still normalises to `/test;x=1`, which is what the existing test asserts.

**Coverage.** I have not touched anything for the 0.0% on new code, because from here I cannot tell whether the `handler` module is missing from the Sonar coverage report or whether the report is not being produced for a fork build at all. Happy to look at the workflow if you point me at where coverage is wired up.

As before I cannot run the Maven build here, so please treat CI and the Sonar gate as the source of truth.

stevehu:
No findings. The latest commit (3c1debc) resolves both remaining points and I found no new correctness issues.

What I verified:

S3776 refactor (matchPath → matchWithoutTrailingSlashes) — behavior-preserving. result != null || config == null || !config.isTrailingSlashFallback() is the exact De Morgan inverse of the old result == null && config != null && config.isTrailingSlashFallback(), and the early returns in the helper mirror the old nesting one for one. Dropping the isInfoEnabled/isWarnEnabled guards is safe: both calls use SLF4J parameterized messages, and the only eager argument is ex.getRequestURI(), a field read. That read happens after normalizeRequestPath returned false, and that method computes normalizedURI before touching any setter — so the warn log still reports the untouched original URI.

indexOf(';', lastIndexOf('/')) hardening — correct, and it fixes more than the unreachable case it targeted:
- /a/;x=1/b/ (count 1) → lastIndexOf('/') = trailing slash, no ; after it → path is the whole URI → /a/;x=1/b. The old code produced /a;x=1/b/.
- /test;x=1/ (param on the last segment, slash after) now normalizes to /test;x=1; the old code returned null and dropped the fallback.
- /test//;x=1 (count 2) still yields /test;x=1 — the existing assertion at HandlerTest.java:322 covers it.
- lastIndexOf returning -1 behaves as a search from 0, so a slash-less URI (*) is unaffected and falls out on the path.length() - count < 1 guard.

Two things I checked against Undertow 2.4.1 source rather than assuming:
- getRequestURI() excludes the query string. The ; branch of Connectors.setExchangeRequestPath does call setRequestURI(encodedPath) with the full target, but URLUtils.PATH_PARAM_PARSER is constructed with parseUntilSeparator = true and stops at ? or / (URLUtils.java:277), so the loop reaches the ? branch and overwrites the URI with encodedPath.substring(0, i). Without that, lastIndexOf('/') could have landed inside a query like ?a=b/ and trimmed it.
- The trim count can never exceed the path's slashes: trimTrailingSlashes always leaves at least one character, so requestPath.length() - trimmedPath.length() is bounded correctly even for // or http://host//.

Encoded-slash safety still holds. The per-character path.charAt(...) != '/' loop is what guards it — /test%2F with DECODE_SLASH on gives a request path of /test/ but a raw URI ending in F, so the fallback is still dropped and the request falls to defaultHandlers.

Tests: I ran them this time — mvn -o -pl handler test -Dtest=HandlerTest → 14/14 pass. The refactor needed no test edits, as the PR author said.

One non-finding, and it's on me since I asked for it: LOG.info fires once per fallback-matched request. On a gateway with trailingSlashFallback enabled and a client that consistently sends trailing slashes, that's an INFO line per request. If that turns out to be noisy in practice, the fix is to drop it to DEBUG and keep the warn — the warn is the one that signals something actionable. Not worth blocking the merge over.

Separately, the Sonar "0.0% coverage on new code" is still open and, as the author notes, it's a CI wiring question rather than a PR change — the gate passes, but the handler module's JaCoCo report evidently isn't reaching Sonar for fork builds.

PR #2788, opened 2026-09-13
partially fixes #13 mask a body that is already parsed into a map or a list

### LINKED ISSUE #13: Mask body in list or map format instead of string format
We have a mask module to mask sensitive data from header, query parameter, path parameter and body. It works with JSON object with JSON Path with configuration. With the introduction of Body module that parses body of application/json to list or map, we need to update the mask module to work with list and map. Regarding to traverse the list or map, here is an example. 
https://github.com/networknt/light-java/blob/master/sanitizer/src/main/java/com/networknt/sanitizer/SanitizerHandler.java

gohri81-tech:
## Summary

Two gaps that are left in the mask module now that the body module hands over a parsed body.

`Mask.maskObject(Object, String)` masks a body that has already been parsed into a `Map` or a `List` and returns it in the same format, so a caller that works with the parsed body does not have to parse the masked JSON string again. Anything that is not a parsed object or array is returned untouched. The existing `maskJson` overloads are unchanged.

`applyMask` only masked `String`, `Integer` and `List` values and logged `The value specified by path {} cannot be masked` for everything else. A long account number, a decimal amount or a boolean flag was therefore left in the clear in the audit log. `Number` and `Boolean` are now masked as well, which also covers `Long`, `Double` and `BigDecimal`.

## Notes

I see this as a partial fix for the original issue rather than the whole of it, so please treat the closing keyword in the first commit message with that in mind and reopen the issue after merging if there is scope left to track. Happy to widen or narrow what is in here.

Masking a numeric value turns it into a string of asterisks in the output, which is what already happened for `Integer`.

## Tests

Added `testMaskNonStringScalars`, `testMaskParsedMapBody`, `testMaskParsedListBody` and `testMaskObjectLeavesOtherInputAlone` to `MaskTest`, with two new keys in the test `mask.yml`. I was not able to run the Maven build in my environment, so please let CI be the judge on that.


gohri81-tech:
The Sonar gate failure is addressed in a2638da.

The failing condition was the reliability rating, from a single blocker: `javabugs:S2259` ("A NULL object should not be accessed") at `Mask.java:197`, the `return ctx.json();` in the new `maskObject`. The data flow Sonar shows goes through `maskJson(DocumentContext, String)`, which starts with `if(ctx == null) return null;`. Because of that guard the analyser treats the parameter as nullable, assumes the branch can be taken, and therefore holds that `ctx` may be null when it is dereferenced on the following line.

`maskObject` now returns the input unchanged when the parse yields no context:

```java
DocumentContext ctx = JsonPath.parse(input);
if(ctx == null)
    return input;
maskJson(ctx, key);
return ctx.json();
```

That removes the path Sonar was walking. `JsonPath.parse` does not return null for a `Map` or a `List`, so no behaviour changes and the tests are untouched. I went this way rather than dropping the null guard from `maskJson(DocumentContext, String)`, since that method is public and the guard is existing behaviour for its other callers - happy to do it the other way round if you would rather the guard went.

As before I cannot run the Maven build here, so please treat CI and the Sonar gate as the source of truth.

stevehu:
From my evaluation, this mask module is not being used by any known user at the moment and majority of our users are moving to the rust implementation light-fabric/apps/light-gateway. We don't need to update this as there is no immediate requirement. Let me know what do you think? Thanks.

gohri81-tech:
Thanks for taking the time to look, and for merging #2787.

That reasoning makes sense to me. If the mask module has no known users and the effort is going into the Rust light-gateway, I would not want to add maintenance surface here for a caller that may never exist, so I am happy to defer to you.

Two options, whichever is less work for you: close this and leave #13 open for whoever picks the equivalent behaviour up on the Rust side, or park it here in case the module is touched again - it is self contained (one new maskObject overload plus Number and Boolean in applyMask) and the Sonar gate is green, so it should not rot badly. Say the word and I will close it myself.

One small piece worth separating out either way: applyMask currently logs "The value specified by path {} cannot be masked" for Number and Boolean, so a long account number or a decimal amount ends up in the audit log in the clear. That part reads as a small bug rather than a new feature, if you would rather have it on its own later.

gohri81-tech:
Closing this as discussed, since the mask module has no known users and the work is moving to the Rust light-gateway. I have left #13 open. If the Java module is ever picked up again, the branch is still on my fork, and the Number/Boolean masking in applyMask is the part I would bring back first. Thanks again for the review.

Issue #2793, opened 2026-09-17
support basePath for k8s ingres in direct-registry.yml


stevehu:
https://github.com/lightapi/light-portal-doc/issues/82
https://github.com/networknt/light-fabric/issues/398

In the light-4j and light-fabric, we have a component called direct-registry and a config file called direct-registry.yml. 

Within this configuration file, we are using a serviceId and envTag (optional) to lookup a host:port for the destination. 

This module is a temp solution before we can do service discovery from the controller-rs. If the destination is in VM, we don't have problem. For example, https://192.168.1.142:8440 or https://api.example.com

However, when the service is deployed in k8s cluster behind an ingress service, the namespace and service must be in the url as path prefix. For example, the target might be https://api.example.com/namespace1/service1. The ingres service will based on the namespace1 and service1 to route the traffic to the right pod. Also the ingress service will remove the /namespace1/service1 so that the request reached to the pod is the normal request with application based path. 

I am wondering if both light-4j and light-fabric support this kind of configuration in the direct-registry.yml file. Could you please evaluate both repos to find out? I remember that one repo has been patched to support it but another is not. Do you think this is a good design or do you have any better ideas to support the k8s ingress? Do you have any recommendations?

stevehu:
Short answer: light-fabric supports it and light-4j doesn't. Your memory is right. light-fabric reads the path in the URL as an upstream prefix. light-4j reads the URL but keeps only scheme, host and port, so the prefix is silently lost.

light-fabric: supported, with one gap

- Parsing: parse_router_target (frameworks/light-pingora/src/router.rs:360) keeps url.path() as ProxyTarget.path_prefix. So https://api.example.com/namespace1/service1 becomes address api.example.com:443 with prefix /namespace1/service1.
- Router: apply_uri_rewrite (router.rs:499) adds the prefix in front of the rewritten request path. The test router_falls_back_to_direct_registry_when_portal_registry_is_absent (router.rs:988) checks exactly this case.
- Gateway proxy: apps/light-gateway/src/main.rs:5530 adds the prefix for non-router targets.
- MCP tools: combine_base_and_tool_path (mcp.rs:8370) joins the base path and the tool path. It also avoids adding the prefix twice.
- Token, SPA auth and JWK lookups: they build server_url + uri as a string, so the prefix survives.
[...]

PR #2795, opened 2026-09-17
fixes #2793 support basePath for k8s ingres in direct-registry.yml

stevehu:


stevehu:
Both addressed and pushed to issue2793 (commit 1a2ee7691); PR #2795 now shows it.

Copilot's review comment (LightCluster.java:135) — valid, and it would have broken real requests. new URI(scheme, null, host, port, path, ...) percent-encodes its path argument, but the base path from direct-registry.yml is already encoded, so a configured https://api.example.com/ns/a%20b came back as /ns/a%2520b from services() — the egress router would then send a prefix the ingress doesn't match. Fixed by building the origin with the component constructor (it still handles IPv6 bracketing) and appending the base path to toASCIIString() verbatim:

URI origin = new URI(url.getProtocol(), null, url.getHost(), url.getPort(), null, null, null);
String basePath = basePath(url);
uri = basePath.isEmpty() ? origin : new URI(origin.toASCIIString() + basePath);

This also makes toUri consistent with serviceToUrl, which already concatenated the base path as a raw string. Cluster.prependBasePath reads getRawPath(), so the encoding survives the join too.

SonarCloud — the quality gate itself is passing (all five conditions OK). What's open is one MINOR code smell: Cluster.java:84, "Remove this hard-coded path-delimiter" (rule S1075). Replaced the "/" literals in prependBasePath with Constants.PATH_SEPARATOR, which is also what the rest of the codebase uses.

Test added — testEncodedBasePathIsNotEncodedTwice in LightClusterTest, backed by a new com.networknt.encoded-1.0.0: https://api.example.com/ns/a%20b entry in the test service.yml. It asserts the raw path stays /ns/a%20b, decodes to /ns/a b, and joins to /ns/a%20b/v1/pets. All seven affected modules build and test clean.

Worth noting: SonarCloud re-analyzes on push, so give the new run a minute before reading the PR page.

stevehu:
Addressed and pushed (246e3e5a5).

Copilot's new finding (DirectRegistry.java:126) is correct. In the legacy service.yml parameter form I was serializing the base path into a synthetic query string and re-parsing it, so /namespace/a&b came back as basePath=/namespace/a plus a junk b parameter, and the request would hit the wrong ingress path.

Rather than escaping the delimiter, I removed the string round-trip that caused it. buildUrl now returns the URL object and sets the parameter on it directly:

URL result = URLImpl.valueOf(u + Constants.PATH_SEPARATOR + key + p);
if(basePath != null) {
    result.addParameter(Constants.BASE_PATH, basePath);
}
return result;

Escaping would have worked for &, but it would leave the base path stored non-verbatim (%26), which contradicts the encoding fix from the previous round — LightCluster concatenates that value straight into the url. Setting it on the object keeps the path exactly as configured, whatever characters it contains. The two call sites lost their now-redundant URLImpl.valueOf(s) step. The direct-registry.yml path never had this bug, since it builds the parameter map directly.

Tests added for both config forms:
- DirectRegistryTest — a parameter entry https://api.example.com/namespace3/a&b?environment=0001 asserts basePath=/namespace3/a&b and environment=0001, so it catches both the truncation and the phantom parameter.
- DirectRegistryConfigTest — the same shape via values.yml, guarding the yml path against a regression.

Both fail against the old code. All seven modules build and test clean.

stevehu:
Addressed and pushed (da47b0a6c).

Copilot's finding (Http2Client.java:1901 and :1927) holds up. callService(protocol, serviceId, …) delegates to callService(URI, …), which calls addHostHeader — that hard-codes Host: localhost whenever the caller didn't set one. An ingress routes on Host and path, so fixing only the path left these calls failing at the ingress. This was pre-existing, but it sits directly in the path this PR is meant to enable.

What I changed:

- Cluster.hostHeader(URI) builds the header from the resolved target, including the port only when it isn't the protocol's default (https://api.example.com:443 → api.example.com, :8443 → api.example.com:8443). Scheme-aware, unlike the old bare 80/443 check.
- Http2Client.applyServiceTarget(URI, ClientRequest) now does both jobs — prepend the base path, then set Host only if the caller hasn't. Both callService serviceId overloads use it, and so do the two ServiceDef constructors in Http2ServiceRequest, which resolve through the cluster the same way and had the identical defect.
- AbstractRemoteMcpTool.buildHostHeader delegates to the shared helper instead of keeping its own near-duplicate.

I scoped the change to the serviceId-resolved entry points. callService(URI, …) called directly still falls back to localhost, since changing that would alter behavior for every existing caller in a PR about base paths — worth a separate look if you want it.

Tests: new Http2ClientServiceTargetTest in the client module (base path + Host applied, non-default port kept, explicit Host preserved) and a testHostHeader case in LightClusterTest. The client test runs without a server, so it stays fast. All seven modules green.

stevehu:
Quality gate: green. It went ERROR → OK after 3730e59fa, with 0 open issues. The two S2259 NPE bugs were self-inflicted: the uri == null || request == null guard I added in applyServiceTarget told Sonar those arguments were nullable, so the existing dereferences in addHostHeader and processClientRequest became reachable-NPE findings. Removing the guards was the right fix — every caller builds the request before applying the target. Also dropped the redundant public modifiers on the new test class (S5786).

New Copilot comment (Http2ServiceRequest.java:108 and :123) — also correct, and a real bug I introduced. The ServiceDef constructors set Host at construction time, but setRequestHeaders(Map) uses add, not put. A caller passing a Host afterwards would get two Host headers, with the generated one first — an invalid request. Fixed by splitting the two concerns (afaf42953):

- Constructors prepend the base path only. That's deterministic and applied exactly once.
- processClientRequest resolves Host at send time, after caller headers are configured. It reads getFirst(Headers.HOST) and only fills in a value when none is set, so a caller-supplied Host wins and is never appended to.

That block already existed ("Ensure host header exists"), so Host resolution is back where it belongs rather than being duplicated in the constructor.

One behavior change worth flagging: that block used hostURI.getHost(), dropping the port. It now uses Cluster.hostHeader, so a non-default port is included — http://localhost:7080 sends Host: localhost:7080 instead of localhost. That's RFC-correct and needed for ingress routing, but it affects every Http2ServiceRequest, not just ServiceDef ones. Say the word if you'd rather scope it to the ServiceDef path.

Tests: two cases in Http2ServiceRequestTest — the derived header (localhost:7080, exactly one value) and a caller-configured Host (petstore.example.com preserved, still one value). processClientRequest went from private to package-private so the test can drive it without a network call. All seven modules pass.

